PR #42, opened 2019-10-10
FIX: Replace imports from ``nipype.workflows`` with new niflow package

oesteban:
This PR is to check whether these changes would work, and releasing from the fix/niflows branch a patch version.

Then, I'll fix the conflicts and update master.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,8 @@ setup_requires =
     setuptools >=40.8
 install_requires =
     nibabel >=2.2.1
-    nipype >=1.2.0
+    niflow-nipype1-workflows == 0.0.1
+    nipype >=1.3.0rc1
     niworkflows ~=0.10
     numpy
     pybids ~=0.9.2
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -24,7 +24,8 @@ setup_requires =
     setuptools >=40.8
 install_requires =
     nibabel >=2.2.1
-    nipype >=1.2.0
+    niflow-nipype1-workflows == 0.0.1
+    nipype >=1.3.0rc1
```
effigies, line 28: This is not a required change. Might go with:

```suggestion
    nipype >=1.2.0,!=1.2.3
```

Unless you explicitly need something from the RC?

oesteban: I believe the etelemetry pacification is found only in 1.3.0rc1 - correct?

effigies: Sure, but etelemetry was only added in 1.2.3.

oesteban: Yup - https://circleci.com/gh/poldracklab/sdcflows/271

effigies:
We already did this with #39... Is there a need to merge? If you're expecting more bug-fixes in 0.1.x, then you can `merge --ours` to avoid conflicts.

```Shell
git fetch upstream
git checkout upstream/master
git merge --ours 0.1.2
git push upstream HEAD:master
```

oesteban:
Oh, I see. I'll just update the nipype pin.

PR #50, opened 2019-11-18
ENH: Stop using siemens2rads from old nipype workflows

oesteban:
A new interface is proposed, to be used in phase-difference based
workflows.

@AKSoo identified a potential issue with siemens2rads
(https://github.com/poldracklab/sdcflows/pull/30#discussion_r346640162),
although the re-centering did not seem to make a big difference.

This implementation of the rescaling uses percentiles to robustly
calculate the range of the input image (as recommended in the original
paper about FSL PRELUDE), and makes sure the output data type is
adequate (float32).

### DIFF
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -566,7 +566,6 @@ def au2rads(in_file, newpath=None):
     data[data > 2.0 * np.pi] = 2.0 * np.pi
     hdr.set_data_dtype(np.float32)
     hdr.set_xyzt_units('mm')
-    hdr['datatype'] = 16
 
     out_file = fname_presuffix(in_file, suffix='_rads', newpath=newpath)
     nb.Nifti1Image(data, im.affine, hdr).to_filename(out_file)
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -574,10 +574,10 @@ def au2rads(in_file, newpath=None):
     data -= mode(data, axis=None)[0][0]
 
     # Scale lower tail
-    data[data < 0] = - np.pi * data[data < 0] / np.percentile(data[data < 0], 2)
+    data[data < 0] = - np.pi * data[data < 0] / data[data < 0].min()
 
     # Scale upper tail
-    data[data > 0] = np.pi * data[data > 0] / np.percentile(data[data > 0], 98)
+    data[data > 0] = np.pi * data[data > 0] / data[data > 0].max()
 
     # Offset to 0 - 2pi
     data += np.pi
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -557,7 +557,7 @@ def _delta_te(in_values, te1=None, te2=None):
 def au2rads(in_file, newpath=None):
     """Convert the input phase difference map in arbitrary units (a.u.) to rads."""
     im = nb.load(in_file)
-    data = im.get_fdata().astype('float32')
+    data = im.get_fdata(dtype='float32')
     hdr = im.header.copy()
 
     data -= np.percentile(data, 2)
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -187,7 +187,16 @@ class _Phasediff2FieldmapOutputSpec(TraitedSpec):
 
 
 class Phasediff2Fieldmap(SimpleInterface):
-    """Convert a phase difference map into a fieldmap in Hz."""
+    """
+    Convert a phase difference map into a fieldmap in Hz.
+
+    This interface is equivalent to running the following steps:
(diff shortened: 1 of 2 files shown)


sdcflows/interfaces/fmap.py
```diff
@@ -531,3 +552,22 @@ def _delta_te(in_values, te1=None, te2=None):
             'EchoTime2 metadata field not found. Please consult the BIDS specification.')
 
     return abs(float(te2) - float(te1))
+
+
+def au2rads(in_file, newpath=None):
+    """Convert the input phase difference map in arbitrary units (a.u.) to rads."""
+    im = nb.load(in_file)
+    data = im.get_fdata().astype('float32')
```
effigies: Not a big deal for a single volume, but this can avoid casting to float64 before casting to float32.

```suggestion
    data = im.get_fdata(dtype='float32')
```

sdcflows/interfaces/fmap.py
```diff
@@ -531,3 +552,22 @@ def _delta_te(in_values, te1=None, te2=None):
...
+    hdr = im.header.copy()
+
+    data -= np.percentile(data, 2)
+    data[data < 0] = 0
+    data = 2.0 * np.pi * data / np.percentile(data, 98)
+    data[data > 2.0 * np.pi] = 2.0 * np.pi
+    hdr.set_data_dtype(np.float32)
+    hdr.set_xyzt_units('mm')
+    hdr['datatype'] = 16
```
effigies: This is a duplicate of `hdr.set_data_dtype(np.float32)` for NIfTI-1/2.

oesteban:
Okay, now I remember why we were not using the 0-6.28 range: it needs recentering (which is not the case for the 2 phases case).

I'll add a recentering node to this PR.

oesteban:
Still doing something odd: https://102-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/fmap/sub-HCP101006_phasediff.svg

kimsin98:
Trimming wrapped phase images at 2nd and 98th percentiles seems odd, since the values outside that range are not really outliers in a wrapped image. They are the connecting values necessary for smooth unwrapping.

In the [original PRELUDE paper](https://doi.org/10.1002/mrm.10354), the percentile trimming is done on the *magnitude* image to derive a mask if no mask is specified by the user.

oesteban:
> Trimming wrapped phase images at 2nd and 98th percentiles seems odd, since the values outside that range are not really outliers in a wrapped image. They are the connecting values necessary for smooth unwrapping.

The intent of this was to deal correctly with phase images cast to float datatypes. Since it doesn't seem to solve any problem, and the particular change goes beyond the scope of this PR, I think you are right I should remove it 👍(done in the latest commit).

sdcflows/interfaces/fmap.py
```diff
@@ -531,3 +561,32 @@ def _delta_te(in_values, te1=None, te2=None):
             'EchoTime2 metadata field not found. Please consult the BIDS specification.')
 
     return abs(float(te2) - float(te1))
+
+
+def au2rads(in_file, newpath=None):
+    """Convert the input phase difference map in arbitrary units (a.u.) to rads."""
+    from scipy.stats import mode
+    im = nb.load(in_file)
+    data = im.get_fdata(dtype='float32')
```
effigies, line 570: Now I'm thinking about this again, I wonder if it makes more sense to make it float64 for the calculation, given that it's only one volume. This would ensure that the rounding error of casting to float32 on write (the `set_data_dtype` below will still take care of that) dominates the arithmetic rounding errors.

oesteban: Feels like splitting hairs... especially for images that originally were int16, DYT?

effigies: Yeah, probably. But it's cheap and reduces accumulated error. That accumulated error is almost certainly in the noise, but the extra guard bits also give you lots of room to have suboptimal order of operations without changing the output relative to an optimal (smallest induced error) algorithm, which feels useful for reproducibility/comparability with other tools.

sdcflows/interfaces/fmap.py
```diff
@@ -187,7 +187,16 @@ class _Phasediff2FieldmapOutputSpec(TraitedSpec):
...
 
 class Phasediff2Fieldmap(SimpleInterface):
-    """Convert a phase difference map into a fieldmap in Hz."""
+    """
+    Convert a phase difference map into a fieldmap in Hz.
+
+    This interface is equivalent to running the following steps:
+      #. Convert from rad to rad/s
+         (``niflow.nipype1.workflows.dmri.fsl.utils.rads2radsec``)
+      #. FUGUE execution: fsl.FUGUE(save_fmap=True)
```
oesteban, line 196:
```suggestion
      #. FUGUE execution: fsl.FUGUE(save_fmap=True).
```

PR #51, opened 2019-11-18
 ENH: Do not use legacy demean function from old nipype workflows

oesteban:
The new demean function works with the median or the mode (default) for a more robust result (to revise after #50 is merged)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,20 +224,12 @@ jobs:
           paths:
             - /tmp/docker
       - run:
-          name: Set-up a Docker registry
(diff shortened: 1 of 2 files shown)


mattcieslak (commented):
Is there a citation for this? I don't see demeaning suggested in the FSL documentation

oesteban:
If I'm not wrong, the demeaning comes from the historic `epiunwarp.fsl` script. Its a massaging that makes sense anyways (e.g. it improves a bit the two phases result because the distributions are better aligned).

Finally, I believe Jenkinson's 2003 paper would cover this.

PR #63, opened 2019-11-27
FIX: Hard-wire ``MNI152NLin2009cAsym`` as standard space for SDC-SyN

oesteban:
Continuing with #55, this PR re-enables SDC-SyN to work with fMRIPrep.

### DIFF
--- a/sdcflows/workflows/syn.py
+++ b/sdcflows/workflows/syn.py
@@ -23,7 +23,7 @@ Feedback will be enthusiastically received.
 
 
 """
-import pkg_resources as pkgr
+from pkg_resources import resource_filename
 
 from nipype import logging
 from nipype.pipeline import engine as pe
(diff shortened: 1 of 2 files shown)


sdcflows/workflows/syn.py
```diff
@@ -153,6 +151,8 @@ def init_syn_sdc_wf(omp_nthreads, epi_pe=None,
     # 1) BOLD -> T1; 2) MNI -> T1; 3) ATLAS -> MNI
     transform_list = pe.Node(niu.Merge(3), name='transform_list',
                              mem_gb=DEFAULT_MEMORY_MIN_GB)
+    transform_list.inputs.in3 = resource_filename(
+        'sdcflows', 'data/fmap_atlas_2_MNI152NLin2009cAsym_affine.mat')
```
oesteban, line 155: In reality, we only have this pre-calculated affine, so offering to switch was asking for trouble.

effigies (approved):
Sounds good. Without a plan to calculate an alternative, or bundling one for each template in templateflow, it doesn't make much sense to enable this flexibility.

We can revisit, if that ever seems like a good idea.

PR #70, opened 2019-11-28
Subtract phases before unwrapping

kimsin98:
Swapped order of subtract/unwrap. Due to how PRELUDE algorithm works, this should create the same results in most cases and produce smoother phasediff map in some cases that are problematic currently.

https://github.com/poldracklab/sdcflows/pull/30#issuecomment-555881469
https://github.com/poldracklab/sdcflows/issues/59


oesteban (approved):
I can't see any difference in the results w.r.t. the baseline implementation, can you @mattcieslak?

I guess that in the absence of differences, this approach will shave off a couple of nodes --one of them being a PRELUDE run, which is not super-costly but we are better off without it.

Issue #73, opened 2019-12-04
PEPOLAR workflow breaks if input images are not 4D


oesteban:
Here https://github.com/poldracklab/sdcflows/blob/6b6ef2a5453388eefc1f97e9e0207dfd16f13649/sdcflows/workflows/pepolar.py#L368

PR #74, opened 2019-12-04
FIX: ValueError when one of the PEPOLAR fielmaps is not 4D

oesteban:
The idea here was to avoid loading the whole image into memory to extract the first `max_trs` volumes at most. @effigies, is this the better way of doing this?

Addresses #73

### DIFF
--- a/sdcflows/workflows/pepolar.py
+++ b/sdcflows/workflows/pepolar.py
@@ -365,7 +365,10 @@ def _split_epi_lists(in_files, pe_dir, max_trs=50):
 
     for i, (epi_path, epi_pe) in enumerate(in_files):
         if epi_pe[0] == pe_dir[0]:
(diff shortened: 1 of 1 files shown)


effigies:
You're loading all of the data in `four_to_three` because you're creating all of the 3D images, even if you immediately drop them afterwards. If you're willing to set a minimum of nibabel 2.3, then you can get `img.slicer`, which will avoid creating those images at all:

```Python
img = nb.load(epi_path)
if len(img.shape) == 3:
    split_nii = [img]
else:
    split_nii = nb.four_to_three(img.slicer[:, :, :, :max_trs])
```

You're not going to get around checking the shape in some way.

effigies:
@oesteban Went ahead and directly proposed #74 to get tests going.

Issue #23, opened 2019-07-22
Too liberal mask of the magnitude image


oesteban:
@zenkavi this happened with your dataset:

![sub-200148_desc-brain_mask](https://user-images.githubusercontent.com/598470/61653326-e206fe80-ac6e-11e9-9787-b01e3c441b01.png)

Although it doesn't really seem to affect registration to the BOLD reference, we want to get this mask right 100% of the times.

zenkavi:
@oesteban thanks for catching. Do you have the subject number? Should I re-run?

oesteban:
> Do you have the subject number?

200148

> Should I re-run?

Nope, it seems to robustly replicate - and it doesn't seem to affect too badly in susceptibility correction. We can live with this.

effigies:
See also [Neurostars #4789](https://neurostars.org/t/fmriprep-fieldmap-skull-stripping-problem/4789).

rob-mccutcheon:
Hi, I have been having problems with the skullstripping of fieldmaps while using fmriprep as detailed in [ Neurostars #4789](https://neurostars.org/t/fmriprep-fieldmap-skull-stripping-problem/4789). 

It was kindly suggested by @effigies to try modifying the sdcflows code, which I imagine might involved changing the value set for 'frac' at line 64[here](https://github.com/poldracklab/sdcflows/blob/master/sdcflows/workflows/fmap.py)


If I were to do this I am unsure, however, as to how then subsequently integrate the change I would make into my fmriprep singularity image? Any advice much appreciated

oesteban:
@AKSoo

kimsin98:
Ah, so BET threshold is a recurring problem.

What would be the best solution though? Just raising the current value (0.6) is a simple option, but that might not be robust. Another simple solution is adding it as an optional parameter. I've checked what parameters various other pipelines used for magnitude brain extraction and seen values ranging from 0.3 to 0.7, so that solution might actually be valid.

A hard option is implementing a more complicated but robust protocol. Perhaps adapting niworkflows skullstrip workflow?

@robmcc10 Late but you can patch docker/singularity images by mounting directories/files into them, as seen in the [fMRIPrep docs](https://fmriprep.readthedocs.io/en/stable/contributors.html#patching-working-repositories).

PR #76, opened 2019-12-13
SDC-pepolar: topup implementation

markushs:
Here's a solution that works for us (ref https://github.com/poldracklab/sdcflows/issues/37).

`topup.py` is a modification of `pepolar.py` and reuses many of the functions/workflows found in `pepolar.py`.

Note, I had to decrease the bet threshold in `gre.py` as the epis (which are used as magnitude inputs to `fmap_wf`) already got skullstripped in `topup/init_prepare_epi_wf`. This is probably not the best solution, as the relevant BET node is also used by other workflows with non-skullstripped mag-images. However I see that there's an ongoing discussion about changing the approach to something more robust than BET (https://github.com/poldracklab/sdcflows/issues/23).

### DIFF
--- a/sdcflows/workflows/base.py
+++ b/sdcflows/workflows/base.py
@@ -170,7 +170,7 @@ was calculated for a more accurate co-registration with the anatomical reference
 
         workflow.connect([
             (inputnode, topup_wf, [
-                ('epi_brain', 'inputnode.in_reference_brain')]),
+                ('epi_file', 'inputnode.in_reference')]),
             (topup_wf, fmap_wf, [
                 ('outputnode.fieldmap', 'inputnode.fieldmap'),
                 ('outputnode.magnitude', 'inputnode.magnitude')]),
--- a/sdcflows/workflows/base.py
+++ b/sdcflows/workflows/base.py
@@ -112,15 +112,23 @@ accurate co-registration with the anatomical reference.
 
     only_syn = 'syn' in fmaps and len(fmaps) == 1
 
-    # PEPOLAR path
+    # PEPOLAR path (now topup)
     if 'epi' in fmaps:
-        from .pepolar import init_pepolar_unwarp_wf, check_pes
+        from .topup import init_topup_wf, check_pes
+        from .fmap import init_fmap_wf, init_fmap2field_wf
+        from .unwarp import init_sdc_unwarp_wf
 
         # SyN works without this metadata
         if epi_meta.get('PhaseEncodingDirection') is None:
             raise ValueError(
                 'PhaseEncodingDirection is not defined within the metadata retrieved '
                 'for the intended EPI (DWI, BOLD, or SBRef) run.')
+
+        if epi_meta.get('TotalReadoutTime') is None:
+            raise ValueError(
+                'TotalReadoutTime is not defined within the metadata retrieved '
+                'for the intended EPI (DWI, BOLD, or SBRef) run.')
+
         outputnode.inputs.method = 'PEB/PEPOLAR (phase-encoding based / PE-POLARity)'
 
         fmaps_epi = [(v[0], v[1].get('PhaseEncodingDirection'))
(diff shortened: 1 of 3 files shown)


mattcieslak:
Thank you for starting this PR! I was wondering if you have run into any left-right flips between the fieldmaps and the epi data. Topup in some versions of FSL assumes that the image orientation is LAS+. Since fmriprep uses RAS+ this might be difficult to spot.

markushs:
@mattcieslak I haven't seen any left-right flips, no. And visually inspecting epi data pre-post topup (i.e. distorted epi input and undistorted epi output) + inspecting headers suggest no flips have occured. Any bullet-proof way to test if something like this has occured?

oesteban:
> I was wondering if you have run into any left-right flips between the fieldmaps and the epi data.

Are you referring to inconsistency between them? i.e., one of them has an x-form header where there is an X-flip and the data are therefore interpreted in reversed way.

> Topup in some versions of FSL assumes that the image orientation is LAS+.

Assumes LAS+ or assumes the same orientation information for both images? The former would be really dangerous, the latter would be just unfortunate but understandable from a historical point of view.

markushs:
I did some additional testing, and with our data it looks like topup works better when epi fieldmaps are not skullstripped. The latest commit to my branch reflects this. As the "magnitude" input to `fmap.py` now contains skull, the frac-threshold in `gre.py` could be reverted to its original value (0.6).

oesteban:
Hi @markushs, is there anything we can do to help you finalize this?

markushs:
Hi @oesteban, have you had a chance to look at the draft? If the approach seems ok I can try to fix the remaining problems (some tests that need to be updated and an issue with the `_desc_fieldmap_bold.svg`, If I recall correctly). The skullstripping could also be refined, but that seems to be work in progress (ref https://github.com/nipreps/sdcflows/issues/23)

sdcflows/workflows/base.py
```diff
@@ -105,22 +105,29 @@ def init_sdc_estimate_wf(fmaps, epi_meta, omp_nthreads=1, debug=False):
...
+was calculated for a more accurate co-registration with the anatomical reference.
 """
 
     only_syn = 'syn' in fmaps and len(fmaps) == 1
 
-    # PEPOLAR path
+    # PEPOLAR path (now topup)
     if 'epi' in fmaps:
-        from .pepolar import init_pepolar_unwarp_wf, check_pes
+        from .topup import init_topup_wf, check_pes
```
oesteban: I don't love the `topup` name of the module. I would prefer this workflow to be within the `pepolar` module.

sdcflows/workflows/base.py
```diff
@@ -105,22 +105,29 @@ def init_sdc_estimate_wf(fmaps, epi_meta, omp_nthreads=1, debug=False):
         return workflow
 
     workflow.__postdesc__ = """\
-Based on the estimated susceptibility distortion, a corrected
-EPI (echo-planar imaging) reference was calculated for a more
-accurate co-registration with the anatomical reference.
+Based on the estimated susceptibility distortion, a corrected EPI (echo-planar imaging) reference
```
oesteban, line 108: Please rebase or merge with current upstream so differences in this file are real.

sdcflows/workflows/base.py
oesteban, line 116: I don't love the `topup` name of the module. I would prefer this workflow to be within the `pepolar` module.

sdcflows/workflows/topup.py
```diff
@@ -0,0 +1,359 @@
...
+        Topup estimated fieldmap (Hz)
+    magnitude : pathlike
+        The ``fmaps_epi`` after unwarping
+    """
+
+    workflow = Workflow(name=name)
+    workflow.__desc__ = """\
+A B0-nonuniformity map (or *fieldmap*) was estimated based on two (or more)
+echo-planar imaging (EPI) references with opposing phase-encoding
+directions, with `Topup`).
```
oesteban, line 98:
```suggestion
echo-planar imaging (EPI) references of opposed phase-encoding
directions with `topup` @topup (FSL {fsl_ver}).
```

We will need to add a topup entry to the bib file, referencing
> J.L.R. Andersson, S. Skare, J. Ashburner. How to correct susceptibility distortions in spin-echo echo-planar images: application to diffusion tensor imaging. NeuroImage, 20(2):870-888, 2003.

sdcflows/workflows/topup.py
```diff
@@ -0,0 +1,359 @@
...
+            ('out', 'readout_times')]),
+        (topup, outputnode, [
+            ('out_field', 'fieldmap'),
+            ('out_corrected', 'magnitude')]),
+    ])
+
+    return workflow
+
+
+def init_prepare_epi_wf(omp_nthreads, matched_pe=False,
```
oesteban, line 152: Can you unpack for me how much of this workflow overlaps the existing analog?

sdcflows/workflows/topup.py
```diff
@@ -0,0 +1,359 @@
...
+    return inlist
+
+
+def _last(inlist):
+    if isinstance(inlist, (list, tuple)):
+        return inlist[-1]
+    return inlist
+
+
+def check_pes(epi_fmaps, pe_dir):
```
oesteban, line 290: Ditto

sdcflows/workflows/topup.py
```diff
@@ -0,0 +1,359 @@
...
+    if not opposed_pe:
+        raise ValueError("""\
+None of the discovered fieldmaps has the right phase encoding direction. \
+This is possibly a problem with metadata. If not, rerun with \
+``--ignore fieldmaps`` to skip the distortion correction step.""")
+
+    return matched_pe, matched_pe_dir, opposed_pe_dir
+
+
+def _split_epi_lists(in_files, pe_dir, max_trs=50):
```
oesteban, line 313: Ditto

sdcflows/workflows/base.py
```diff
@@ -131,21 +138,54 @@ def init_sdc_estimate_wf(fmaps, epi_meta, omp_nthreads=1, debug=False):
...
+
+        fmap2field_wf = init_fmap2field_wf(
+            omp_nthreads=omp_nthreads,
+            debug=debug)
+        # set inputs
+        fmap2field_wf.inputs.inputnode.metadata = epi_meta
+
+        sdc_unwarp_wf = init_sdc_unwarp_wf(
+            omp_nthreads=omp_nthreads,
+            debug=debug)
```
oesteban, line 169: This is not a responsibility of this workflow. Instead, the warpfield should be converted to an ITK-type of transform and generate a new (unwarped) reference, as in: https://github.com/nipreps/sdcflows/blob/bf9cd97fa1e0e25d152c02e96120a1ed37b60a54/sdcflows/workflows/pepolar.py#L128-L140

But that's going to be tricky while https://github.com/poldracklab/nitransforms does not support FSL's bspline parametric fields.

oesteban (commented):
Getting there! Very looking forward to seeing this finalized :)

markushs:
Hi,
Sorry for the delay here (has been crazy times...)
I have some time now to finalize this, if it's still of interest @oesteban?
As I have to re-familiarize myself with the code (including my own) I was wondering if it would be cleaner to open a new PR based on the current master?

sdcflows/workflows/base.py
markushs: As you can see I've opened a new PR (#106 based on the current master). Will respond to your comments here, but the perhaps better to continue the discussion in the new PR? Sorry for the mess, far from fluent in git...

sdcflows/workflows/base.py
markushs: Fixed in #106

sdcflows/workflows/topup.py
markushs: FSL version gets returned in #106, but not sure where I find the bib-file?

sdcflows/workflows/topup.py
markushs: In #106 all edits were done in `pepolar.py`, so should be easier to evaluate the overlap

sdcflows/workflows/topup.py
markushs: Ditto

sdcflows/workflows/topup.py
markushs: Ditto

sdcflows/workflows/base.py
markushs: Not sure what you mean by this not being a responsibility of this workflow. I tried to summarize the approach in the comment to (#106), but essentially the topup-estimated fieldmap is fed into the workflows of the fmap-module.

oesteban:
I'll have a closer look later today or tomorrow. Since this PR seemed a bit abandoned, I made some progress on this front here: https://github.com/nipreps/dmriprep/pull/97/files. Hopefully, we can find the common points of that PR to dmriprep and this one to come up with a solution that works well regardless of the input dataset.

oesteban:
Let's move the conversation to #106.

Issue #77, opened 2019-12-17
Spiral fieldmap estimation workflow is broken


oesteban:
@poldrack just hit this wall
```
Node: fmriprep_wf.single_subject_s573_wf.func_preproc_ses_1_task_DPX_run_1_wf.sdc_estimate_wf.fmap_wf.fmapmrg
Working directory: /scratch1/01329/poldrack/uh2_workdir/fmriprep_wf/single_subject_s573_wf/func_preproc_ses_1_task_DPX_run_1_wf/sdc_estimate_wf/fmap_wf/fmapmrg

Node inputs:

hmc = False
in_files = ['/scratch1/01329/poldrack/BIDS_scans/sub-s573/ses-1/fmap/sub-s573_ses-1_fieldmap.nii.gz']
to_ras = True
zero_based_avg = False

Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 69, in run_node
    result['result'] = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 479, in run
    result = self._run_interface(execute=True)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 585, in _run_interface
    return self._run_command(execute)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 678, in _run_command
    result = self._interface.run(cwd=outdir)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 382, in run
    runtime = self._run_interface(runtime)
  File "/usr/local/miniconda/lib/python3.7/site-packages/niworkflows/interfaces/images.py", line 83, in _run_interface
    hmcdat = hmcnii.get_fdata().mean(axis=3)
  File "/usr/local/miniconda/lib/python3.7/site-packages/numpy/core/_methods.py", line 62, in _mean
    rcount = _count_reduce_items(arr, axis)
  File "/usr/local/miniconda/lib/python3.7/site-packages/numpy/core/_methods.py", line 55, in _count_reduce_items
    items *= arr.shape[ax]
IndexError: tuple index out of range
```

[...]

PR #78, opened 2019-12-17
pin: latest niworkflows

mgxd:
spotted when building fMRIPrep docker image:

```
ERROR: sdcflows 1.0.1 has requirement niworkflows~=1.0.0, but you'll have niworkflows 1.1.0 which is incompatible.
```


oesteban:
Let's wait for #77 since it needs niworkflows updates.

PR #81, opened 2020-01-27
FIX: Connect SyN outputs whenever SyN is run

effigies:
Ref: https://neurostars.org/t/force-syn-error-fmriprep-1-5-7/5871


effigies:
I don't have write access to @mgxd's branch. Replaced by #82.

PR #82, opened 2020-01-27
FIX: Connect SyN outputs whenever SyN is run

effigies:
Replaces #81.

Ref: https://neurostars.org/t/force-syn-error-fmriprep-1-5-7/5871

### DIFF
--- a/sdcflows/workflows/base.py
+++ b/sdcflows/workflows/base.py
@@ -233,6 +233,8 @@ accurate co-registration with the anatomical reference.
                 ('epi_brain', 'inputnode.in_reference_brain'),
                 ('t1w_brain', 'inputnode.t1w_brain'),
                 ('std2anat_xfm', 'inputnode.std2anat_xfm')]),
+            (syn_sdc_wf, outputnode, [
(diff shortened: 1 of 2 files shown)


sdcflows/workflows/base.py
```diff
@@ -287,6 +285,10 @@ def fieldmap_wrangler(layout, target_image, use_syn=False, force_syn=False):
...
             })
 
-    if force_syn is True or (not fieldmaps and use_syn is True):
-        fieldmaps['syn'] = force_syn
+    if fieldmaps and force_syn:
+        # syn: True -> Run SyN in addition to fieldmap-based SDC
+        fieldmaps['syn'] = True
+    elif not fieldmaps and (force_syn or use_syn):
+        # syn: False -> Run SyN as only SDC
+        fieldmaps['syn'] = False
```
effigies, line 293: @oesteban This is a change in logic that you should review. IMO, if there are no fieldmaps, we should treat `--force-syn` identical to `--use-syn-sdc`.

oesteban: Like it, it's more explicit.

oesteban (approved):

PR #84, opened 2020-02-03
REL: 1.1.0

effigies:
When installing `fmriprep` from the repository, I get:

```
ERROR: sdcflows 1.0.4 has requirement niworkflows~=1.0.3, but you'll have niworkflows 1.1.4 which is incompatible.
```

To ensure that `~=` constraints keep a compatible batch of releases, I propose that we start a 1.1.x series that depends on `niworkflows ~=1.1.4`.

If bug fixes are needed with 1.0.x, we can create a maintenance branch.

I've also moved the `name` metadata to `setup.py`, which allows GitHub's dependency tree to find sdcflows.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,4 @@
 [metadata]
-name = sdcflows
 url = https://github.com/poldracklab/sdcflows
 author = The SDCflows developers
 author_email = code@oscaresteban.es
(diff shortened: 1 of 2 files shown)


mattcieslak (approved):
Looks good

PR #88, opened 2020-02-13
RF/FIX: Scale all phase maps to ``[0, 2pi]`` range

### LINKED ISSUE #83: Invalid transformations for some Phase1/Phase2 fieldmaps
We encounter an error (log below) in phmap2rads that we have associated with a certain type of fieldmap acquisition but don't know why it fails. Specifically, the when the BOLD scan and the fieldmap scan are both acquired axially there is no error, but when the fieldmap is acquired sagitally the error happens. The error message below is followed by examples of fieldmaps that work and don't work.

========================

/DATA/fmriprep-1.5.5/fmriprep/sub-R/log/20200124-061243_5ea8ae31-a295-496d-a586-748a17830a8f/crash-20200124-062515-fmriprep-phmap2rads-2f64b28a-ad9b-497a-b4a0-1242888fad3c.txt
```
Working Directory: /DATA/fmriprep-1.5.5/work_sub-R/fmriprep_wf/single_subject_R_wf/func_preproc_ses_05_task_taskmotor_run_01_wf/sdc_estimate_wf/phdiff_wf/phmap2rads
Inputs:

    in_file: ['/DATA/fmriprep-1.5.5/bids/sub-R/ses-05/fmap/sub-R_ses-05_run-01_phasediff.nii.gz']

  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/plugins/multiproc.py", line 336, in _send_procs_to_workers
    self.procs[jobid].run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1373, in _run_interface
    stop_first=str2bool(self.config["execution"]["stop_on_first_crash"]),
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 1295, in _collate_results
    "Subnodes of node: %s failed:\n%s" % (self.name, "\n".join(msg))
Exception: Subnodes of node: phmap2rads failed:
Subnode 0 failed
Error: Traceback (most recent call last):
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/utils.py", line 94, in nodelist_runner
    result = node.run(updatehash=updatehash)
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
  File "/usr/local/miniconda/lib/python3.7/site-packages/nipype/interfaces/base/core.py", line 395, in run
  File "/usr/local/miniconda/lib/python3.7/site-packages/sdcflows/interfaces/fmap.py", line 266, in _run_interface
    newpath=runtime.cwd)
  File "/usr/local/miniconda/lib/python3.7/site-packages/sdcflows/interfaces/fmap.py", line 673, in au2rads
    data[data < 0] = - np.pi * data[data < 0] / data[data < 0].min()
  File "/usr/local/miniconda/lib/python3.7/site-packages/numpy/core/_methods.py", line 32, in _amin
    return umr_minimum(a, axis, None, out, keepdims, initial)
ValueError: zero-size array to reduction operation minimum which has no identity
```

When creating this crashfile, the results file corresponding
to the node could not be found.

===================
A fieldmap acquisition that works has the following output from c3d -info:
Image #1: dim = [144, 144, 59];  bb = {[172.569 172.569 -78.3001], [522.569 522.569 80.9999]};  vox = [2.43056, 2.43056, 2.7];  range = [-250, 249.881];  orient = LPI

A fieldmap acquisition that producess the  error above has the following output from c3d -info:
Image #1: dim = [59, 144, 144];  bb = {[78.3001 172.569 -172.569], [237.6 522.569 177.431]};  vox = [2.7, 2.43056, 2.43056];  range = [-250, 249.881];  orient = LPI

effigies:
Final decision: Instead of making `[-pi, pi]` a special case, all data will be naively scaled to `[0, 2pi]`. This ensures that the source data intervals are respected, but may result in a central mode slightly off of `pi`.

We judge that the simplicity and consistency outweighs any small adjustments and potential for bugs resulting from a mode-based transform with separate treatment of the left and right tail.

---

I have some Philips Achieva 2-echo phase/magnitude fieldmaps that come out of `dcm2niix` with a range of approximately `(-pi, pi)`.

This adjusts au2rads to detect that case and simply wraps the negative values to `(pi, 2pi)`.

This may resolve #83.

cc @ins0mniac2

### DIFF
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -666,29 +666,33 @@ def au2rads(in_file, newpath=None):
     data = im.get_fdata(dtype='float32')
     hdr = im.header.copy()
 
-    dmin, dmax = data.min(), data.max()
+    if np.allclose((data.min(), data.max()), (-np.pi, np.pi), atol=0.01):
+        # Already in rads, but wrap to 0 - 2pi
+        data[data < 0] += 2 * np.pi
+    else:
+        dmin, dmax = data.min(), data.max()
 
-    # Find the mode ignoring outliers on the far max/min, to allow for junk outside the FoV
-    fudge = 0.05 * (dmax - dmin)
-    mode = stats.mode(data[(data > dmin + fudge) & (data < dmax - fudge)])[0][0]
+        # Find the mode ignoring outliers on the far max/min, to allow for junk outside the FoV
+        fudge = 0.05 * (dmax - dmin)
+        mode = stats.mode(data[(data > dmin + fudge) & (data < dmax - fudge)])[0][0]
 
-    # Center data around 0.0
-    data -= mode
+        # Center data around 0.0
+        data -= mode
 
-    # Provide a less opaque error if we still can't figure it out
-    neg = data < 0
-    pos = data > 0
-    if not (neg.any() and pos.any()):
-        raise ValueError("Could not find an appropriate mode to center phase values around")
+        # Provide a less opaque error if we still can't figure it out
+        neg = data < 0
+        pos = data > 0
+        if not (neg.any() and pos.any()):
+            raise ValueError("Could not find an appropriate mode to center phase values around")
 
-    # Scale lower tail
-    data[neg] = - np.pi * data[neg] / data[neg].min()
+        # Scale lower tail
+        data[neg] = - np.pi * data[neg] / data[neg].min()
 
-    # Scale upper tail
-    data[pos] = np.pi * data[pos] / data[pos].max()
+        # Scale upper tail
+        data[pos] = np.pi * data[pos] / data[pos].max()
 
-    # Offset to 0 - 2pi
-    data += np.pi
+        # Offset to 0 - 2pi
+        data += np.pi
 
     # Clip
     data = np.clip(data, 0.0, 2 * np.pi)
--- a/sdcflows/interfaces/fmap.py
+++ b/sdcflows/interfaces/fmap.py
@@ -661,46 +661,20 @@ def _delta_te(in_values, te1=None, te2=None):
 
 def au2rads(in_file, newpath=None):
     """Convert the input phase difference map in arbitrary units (a.u.) to rads."""
-    from scipy import stats
     im = nb.load(in_file)
-    data = im.get_fdata(dtype='float32')
+    data = im.get_fdata(caching='unchanged')  # Read as float64 for safety
     hdr = im.header.copy()
 
-    if np.allclose((data.min(), data.max()), (-np.pi, np.pi), atol=0.01):
-        # Already in rads, but wrap to 0 - 2pi
-        data[data < 0] += 2 * np.pi
-    else:
-        dmin, dmax = data.min(), data.max()
+    # Rescale to [0, 2*pi]
+    data = (data - data.min()) * (2 * np.pi / (data.max() - data.min()))
 
-        # Find the mode ignoring outliers on the far max/min, to allow for junk outside the FoV
-        fudge = 0.05 * (dmax - dmin)
-        mode = stats.mode(data[(data > dmin + fudge) & (data < dmax - fudge)])[0][0]
-
-        # Center data around 0.0
-        data -= mode
-
-        # Provide a less opaque error if we still can't figure it out
-        neg = data < 0
-        pos = data > 0
-        if not (neg.any() and pos.any()):
-            raise ValueError("Could not find an appropriate mode to center phase values around")
-
-        # Scale lower tail
-        data[neg] = - np.pi * data[neg] / data[neg].min()
-
-        # Scale upper tail
-        data[pos] = np.pi * data[pos] / data[pos].max()
-
-        # Offset to 0 - 2pi
-        data += np.pi
-
-    # Clip
-    data = np.clip(data, 0.0, 2 * np.pi)
+    # Round to float32 and clip
+    data = np.clip(np.float32(data), 0.0, 2 * np.pi)
 
     hdr.set_data_dtype(np.float32)
     hdr.set_xyzt_units('mm')
     out_file = fname_presuffix(in_file, suffix='_rads', newpath=newpath)
-    nb.Nifti1Image(data, im.affine, hdr).to_filename(out_file)
+    nb.Nifti1Image(data, None, hdr).to_filename(out_file)
     return out_file
 
 


sdcflows/interfaces/fmap.py
```diff
@@ -666,17 +666,21 @@ def au2rads(in_file, newpath=None):
     data = im.get_fdata(dtype='float32')
     hdr = im.header.copy()
 
-    # First center data around 0.0.
-    data -= mode(data, axis=None)[0][0]
+    if np.allclose((data.min(), data.max()), (-np.pi, np.pi), atol=0.01):
+        # Already in rads, but wrap to 0 - 2pi
+        data[data < 0] += 2 * np.pi
```
oesteban: Why only negatives?
```suggestion
        data -= data.min()
```

effigies: Because this function sets the range to (0, 2pi), not (pi, 3pi). And it seems unwise to go precisely counter-phase by adding `np.pi` to everything.

oesteban: Sorry, I had the suggestion above sign-flipped. I've fixed it.

This is going to wrap and scale only negatives - you don't want either (and even less only for the lower half of the histogram).

If I understand correctly, the initial range goes from -pi to pi. Adding pi will move the range to 0-2pi

effigies: So does fieldmap correction treat (θ±π) the same? Shifting by half phase feels problematic.

effigies: To be clear, this is equivalent to:

```
data = (data + 2 * np.pi) % (2 * np.pi)
```

But without bothering add, divide and round things in the (0, pi) range. It doesn't feel controversial to me. `data += np.pi` does.

oesteban: > So does fieldmap correction treat (θ±π) the same? Shifting by half phase feels problematic.

FSL FUGUE theoretically only unwraps phase maps in the (0, 2π) range. It seems to work with (-π, π), though.

I don't know why `data -= np.pi` is controversial - it just moves all values from (-π, π) to (0, 2π).

The transformation you propose will wrap negative phase values (which you definitely don't want to do).

Normally, a phase maps' histogram looks like a centered pyramid. Doing this, we are selecting the left side of the pyramid, shifting it to overlap the right side of the pyramid and finally checking no values have gone over 2pi. That will introduce another wrapping that FUGUE might or might not pick up.

EDIT: The shifted intensities will not overlap, now that I look into this more carefully, but the final distribution will be like a V, with the notch at pi.

effigies: Let's talk in the meeting.

sdcflows/interfaces/fmap.py
```diff
@@ -666,17 +666,21 @@ def au2rads(in_file, newpath=None):
...
 
-    # Scale lower tail
-    data[data < 0] = - np.pi * data[data < 0] / data[data < 0].min()
+        # Scale lower tail
+        data[data < 0] = - np.pi * data[data < 0] / data[data < 0].min()
 
-    # Scale upper tail
-    data[data > 0] = np.pi * data[data > 0] / data[data > 0].max()
+        # Scale upper tail
+        data[data > 0] = np.pi * data[data > 0] / data[data > 0].max()
```
oesteban: ~~This (the old code) looks problematic - we shift the left side of the pyramid first, hence making it overlap with the right hand side. Operations should be done the other way around.~~

Okay, sorry for the noise. There is no shifting yet.

effigies:
Here's my histogram: Figure_1

effigies:
Results of live discussion:

Replace mode-centering and separately rescaling positive and negative to simply scaling min/max to 0/2pi. This would affect fieldmaps from all scanners.

@mattcieslak @AKSoo Your input here would be appreciated.

sdcflows/interfaces/fmap.py
```diff
@@ -663,28 +663,19 @@ def au2rads(in_file, newpath=None):
     """Convert the input phase difference map in arbitrary units (a.u.) to rads."""
     from scipy.stats import mode
     im = nb.load(in_file)
-    data = im.get_fdata(dtype='float32')
+    data = im.get_fdata(caching='unchanged')  # Read as float64 for safety
```
effigies, line 665: I'm extremely comfortable using double memory for one volume to keep things deterministic.

sdcflows/interfaces/fmap.py
```diff
@@ -663,28 +663,19 @@ def au2rads(in_file, newpath=None):
...
-    # Clip
-    data = np.clip(data, 0.0, 2 * np.pi)
+    # Round to float32 and clip
+    data = np.clip(np.float32(data), 0.0, 2 * np.pi)
 
     hdr.set_data_dtype(np.float32)
     hdr.set_xyzt_units('mm')
     out_file = fname_presuffix(in_file, suffix='_rads', newpath=newpath)
-    nb.Nifti1Image(data, im.affine, hdr).to_filename(out_file)
+    nb.Nifti1Image(data, None, hdr).to_filename(out_file)
```
effigies, line 677: If you're keeping the affines in the header, there's no need to pass an affine in to override.

sdcflows/interfaces/fmap.py
```diff
@@ -663,32 +663,19 @@ def au2rads(in_file, newpath=None):
...
```
oesteban, line 677: ~~Why is the affine dropped?~~

effigies (commented):
Just a couple explainers before you get to asking. :-)

oesteban (approved):

kimsin98:
> Replace mode-centering and separately rescaling positive and negative to simply scaling min/max to 0/2pi. This would affect fieldmaps from all scanners.

I also had doubts about mode centering. This looks simpler and cleaner.

effigies:
@mgxd I see the CI has gotten clogged. Naturally.

If it makes life easier, I can cancel this and master and just merge this to only have one job in queue.

mgxd:
Doesn't the current master CI needs to pass to upload the new version to PyPI?

I'm fine with cutting corners, but would like to see a green check here prior to merging just for my peace of mind

effigies:
Okay. I'll leave this one up. Still cancelled master since I rebased this on master.

We'll push to PyPI with the tag: https://circleci.com/workflow-run/52c4f144-141b-41ed-acad-e218e1b65f2a

PR #90, opened 2020-02-14
CI: Fix package tests

effigies:
`test_package` is failing. Might as well fix it while we wait on nipype release.


effigies:
Looks like they require an updated nipype to run.

PR #91, opened 2020-02-14
FIX: Circle pypi deployment

mgxd:


.circleci/config.yml
```diff
@@ -442,32 +442,35 @@ jobs:
...
           command: |
-            virtualenv --python=python3 /tmp/install
-            source /tmp/install/bin/activate
+            virtualenv --python=python3 /tmp/install_sdist
+            source /tmp/install_sdist/bin/activate
             python3 -m pip install "setuptools>=30.3.0" "pip>=10.0.1"
+            deactivate
+            virtualenv --python=python3 /tmp/install_wheel
+            source /tmp/install_wheel/bin/activate
+            python3 -m pip install "setuptools>30.3.0" "pip>=10.0.1"
```
effigies: If you're not going to test install the wheel (and I'm not sure why you should, given that we already do `test_package`), this second environment doesn't need to be made.

But it would probably be good to change the `twine check` lines to check both the sdist and the wheel:

https://github.com/nipreps/sdcflows/blob/1e86d82fa024727b8e614385ef3c87e8d869037a/.circleci/config.yml#L413

https://github.com/nipreps/sdcflows/blob/1e86d82fa024727b8e614385ef3c87e8d869037a/.circleci/config.yml#L480

They can just become `twine check dist/sdcflows*`.

PR #95, opened 2020-03-30
MAINT: Flexibilize dependencies -- nipype, niworkflows, pybids

oesteban:
These changes will make it easier to implement the table in poldracklab/fmriprep#2054.

It also curates a bit the package metadata, now that I was at it.

References: poldracklab/fmriprep#2054

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ install_requires =
     nibabel >=3.0.1
     niflow-nipype1-workflows ~= 0.0.1
     nipype >=1.3.1,<2.0
-    niworkflows >= 1.1.4
+    niworkflows >=1.1.4,<1.3
     numpy
     pybids >= 0.9.2
     templateflow >=0.4
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,13 +1,6 @@
 [metadata]
-url = https://github.com/nipreps/sdcflows
 author = The SDCflows developers
 author_email = nipreps@gmail.com
-maintainer = Oscar Esteban
-maintainer_email = nipreps@gmail.com
-description = Susceptibility Distortion Correction (SDC) workflows for EPI MR schemes.
-long_description = file:README.rst
-long_description_content_type = text/x-rst; charset=UTF-8
-license = Apache-2.0
 classifiers =
     Development Status :: 2 - Pre-Alpha
     Intended Audience :: Science/Research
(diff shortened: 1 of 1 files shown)


oesteban:
Anything against this?

effigies:
Why are you moving from `~=` to `>=`? That seems to be inviting breakage.

oesteban:
I guess the concern is for `niworkflows` and `pybids` in particular (the others already have `>=`).

For `pybids` we already know SDCFlows should work with both >=0.9.2 and 0.10.x. We don't know yet for 0.11, but probably yes (only for SDCFlows). So I think it is okay to leave it open.

For `niworkflows` it is perhaps more complicated, but I don't think there is any change in the horizon that will make it incompatible. Similarly, we could set an extra `,<1.3` as a failsafe.

What do you think?

oesteban:
The above doesn't mean I don't think we should stick with `~=` for fmriprep (and probably for smriprep too), but these dependencies should be as open as possible to minimize the friction in version maintenance.

effigies:
The point of moving to `~=` was to put some brakes on the changes that could be inflicted on slower-moving downstream projects, so moving to open `>=` is I think a serious regression.

But I would be okay with moving from `~=1.1.4` (equivalent to  `>=1.1.4,<1.2`) to `>=1.1.4,<1.3`. And then if we remain compatible at 1.3, we can eventually bump to `<1.4`. This is a sensible process to force us to make sure that we are still compatible and pop out a patch release with an updated version range.

effigies:
Incidentally, are we testing minimum versions?

oesteban:
> Incidentally, are we testing minimum versions?

I don't think so.



> The point of moving to `~=` was to put some brakes on the changes that could be inflicted on slower-moving downstream projects

Agreed, but I think we need to look at the context. It was necessary for niworkflows, as we were working more on a extend-nipype basis rather than implement-and-upstream. That made the dependency from fmriprep really picky. Then we moved to the implement-and-upstream mode (which is happening very very slowly). We also introduced the split of SDCFlows and sMRIPrep, effectively adding one more layer between niworkflows and fmriprep. We increased the complexity but also separated out the implementation needs. In other words, by being strict on how we pin SDCFlows (`~=`), the breaks for niworkflows can be estimated further away.

Then, for pybids, I honestly think it is stable enough to consider it as just another dependency and we are not putting breaks on those because they don't seem likely to create a conflict.

effigies:
I'm not sure. Are you arguing against setting maximum versions?

oesteban:
Only for niworkflows and pybids. Or at least, against immediate (next minor) maximum versions.

effigies:
So let's do `>=1.1.4,<1.3` for now.

As to PyBIDS, sure, why not.

effigies (approved):

PR #96, opened 2020-03-30
CI: Test minimum dependencies

effigies:
This sets up Travis to test minimum dependencies. I don't have permissions to turn on Travis for SDCflows.


effigies:
@oesteban This is failing with nipype 1.3.1 and traits 6.0.0. The issue was fixed in 1.4.2 (https://github.com/nipy/nipype/pull/3159).

Not sure what the best way to handle sudden API breakages in packages a couple levels upstream would be. It might be that testing minimum dependencies in a package with this big a dependency tree is just not feasible.

oesteban:
> Not sure what the best way to handle sudden API breakages in packages a couple levels upstream would be. It might be that testing minimum dependencies in a package with this big a dependency tree is just not feasible.

Yeah, I don't have an answer to that either. What do you feel would be the best course of action on this matter, at this point?

oesteban:
This PR, amongst other things, has triggered the combo d01d88229ea535beeebfa8f76fb29b802f08783b, 01c8188735ffa76f870997b47a27718625f4b143 (which means, has made me revise all the dependencies) - so it is a good thing to merge asap.

Iterating over this, we can try to expand this base to follow up dependencies of niworkflows, nipype, pybids and templateflow. Including something like that, we should be in acceptable shape.

PR #100, opened 2020-04-16
FIX: WM mask selection from dseg before generating report

### LINKED ISSUE #99: Unwarping report: use BIDS-Derivatives-compatible label for WM
As a result of addressing poldracklab/smriprep#176, these two lines:

https://github.com/nipreps/sdcflows/blob/8f25fd22157cf0477d4d8151d158850737f09570/sdcflows/workflows/outputs.py#L59-L60

Should be updated as in poldracklab/fmriprep#2084

oesteban:
Following suit of poldracklab/fmriprep#2084 and inline with nipreps/niworkflows#500, this PR uses a more robust version of the extraction code.

Resolves: #99
Depends: nipreps/niworkflows#500

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,7 +3,7 @@ git+https://github.com/rwblair/sphinxcontrib-versioning.git@39b40b0b84bf872fc398
 nbsphinx
 niflow-nipype1-workflows ~= 0.0.1
 nipype>=1.3.1
-niworkflows ~= 1.1.2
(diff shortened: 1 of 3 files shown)


effigies:
Does this have any impact on derivatives?

oesteban:
Nope, only on the reports. But we may want to downstream it to 1.1.x

effigies:
sdcflows 1.1.x? I don't think anything depends on it.

oesteban:
fair enough - why bother 👍

PR #103, opened 2020-05-04
FIX: backport niworkflows utility function

mgxd:
https://github.com/poldracklab/fmriprep/pull/2115#issuecomment-622550772

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,7 @@
+1.2.4 (May 4, 2020)
+===================
+Bug-fix release to ensure backwards compatibility with previous fMRIPrep versions.
+
 1.2.3 (April 16, 2020)
 ======================
 Bug-fix release enforcing BIDS-Derivatives labels on ``dseg`` file.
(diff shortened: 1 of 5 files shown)


oesteban:
Wouldn't this backport pertain to niworkflows?

mgxd:
We could go about it either way? This seems like the easier route IMO, and should be replaced once we go bump to `1.3.0`

oesteban:
I believe this function is also used in smriprep and I think that we will need to also backport there if we want to be able to progress after 0.5.2 in the 0.5.x series.

mgxd:
shouldn't be necessary as smriprep 0.5.x series is pinned to niworkflows 1.1.x. (all these different versions are making our dependency matrix more complex than necessary, which led to this proposal in https://github.com/poldracklab/fmriprep/issues/2054#issuecomment-623509494)

mgxd:
closing and releasing from the `maint/1.2.x` branch.

no need to include this in master.

PR #108, opened 2020-08-12
ENH: Return out_warp == "identity" if no SDC is applied

effigies:
This change allows the `out_warp` output of `init_sdc_estimate_wf` to be used regardless of whether SDC was run. `antsApplyTransforms` accepts values of `"identity"`in a transform chain.


effigies:
Merging to smooth the path for tomorrow.

@oesteban @mattcieslak Please shout if you have concerns.

Issue #26, opened 2019-07-24
ENH: Optional workflow to write warpings out to ``derivatives/sub-ID/fmap/``


mattcieslak:
Would it make sense to add the SDC warp in the derivatives for each unwarped image? At the moment we are saving the entire working directory to keep them. They're also not included in the h5 transform to the template because they differ for each scan. What do you think?

effigies:
In general, I would say if people are digging through the working directory to find something that's not easily generated from other derivatives, it's probably worth having as a derivative.

oesteban:
This should be covered with the super-fancy X5 format. Sadly I am very slow with the implementation :(

effigies:
But we do have the warps, right? I don't think we need to have the final version ready before we provide the derivatives.

oesteban:
Let's make an optional hook to the fieldmap estimation workflows to write out these files while we build the X5 transforms file.

oesteban:
Okay, I'd like to take my previous comments back and push for this. However, I'd like to modify the original intent of @mattcieslak a little bit:

Under the `fmap/` folder you'd find an actual field map in Hertz, after all the curation process (i.e., estimation, smoothing/cleanup, and extrapolation). So we are not talking about warpings - just the field inhomogeneity map.

I would extend this issue to sending a PR to the imaging-general-derivatives in BIDS to describe the naming of this fieldmap (we could keep the fieldmap suffix, I believe), and enforce the units and allowing for some "IntendedFor" field? (will need to think more about this).

Calculating warpings then should be fairly straightforward: you have the field in Hz and you need to read the EffectiveEchoSpacing and PhaseEncodingDirection of your EPI dataset(s) to properly calculate the warp field.

Issue #21, opened 2019-05-28
SDC: Unify unwarping workflows


oesteban:
After estimation of the displacements field map (either via phase1/2, phasediff, PE-polar, or SyN), the unwarping should happen with a unique workflow.

This requires #19.

oesteban:
Very related: #26

oesteban:
Actually, this is essentially implementing #26. Closing in favor of that one.

PR #113, opened 2020-10-14
ENH: Refactor the workflow infrastructure

oesteban:
This PR essentially creates a new ``init_fmap_preproc_wf`` workflow factory function, inspired by the success of ``init_anat_preproc_wf`` in *sMRIPrep* for T1w, T2w, and FLAIR images.

The idea is to clearly split the SDC (susceptibility distortion correction) process into its two main steps (#21): estimation of the warping and *unwarping* of data. To do so, fieldmap data are regarded at a higher level that actually generates derivatives (ref. #26).

The new `init_fmap_preproc_wf` takes in some ``input_data`` and:

1) Determines what kind of fieldmap estimation strategy requires
2) Stages the correct workflow for estimation
3) Writes out the estimated and preprocessed (e.g., smoothing/despiking) fieldmap with mm/s units. This means that the actual distortion field can be generated by scaling the fieldmap with the Effective echo spacing (in seg), along the appropriate PE direction (and with the appropriate sign) from the PhaseEncodingDirection field of the target dataset.

Because there's no spec for fieldmap derivatives, we could store it as a 4D file with two timepoints. First timepoint is the actual fieldmap and the second is the corresponding anatomical reference (magnitude, epi).

The derivative file would be something like: `derivatives/fmriprep-20.2.0/sub-01/fmap/sub-01_desc-preproc_fieldmap.nii.gz`. Should bids-standard/bids-specification#622 be merged, then a potential name for this could be `derivatives/fmriprep-20.2.0/sub-01/fmap/sub-01_[id-<B0FieldIdentifier>]_fieldmap.nii.gz` with `B0FieldIdentifier` being the metadata entry, if present.

The corresponding JSON metadata should always propagate the `IntendedFor` of the source data, so that these fieldmaps can be actually applied (esp. if bids-standard/bids-specification#622 is never accepted.)

Another feature I'm seriously considering is to create a lightweight object (á la `niworkflows.utils.spaces.Reference`) for representing `input_data` argument.

Some sort of `FieldmapSources` object that morphs itself depending on the inputs, to offer specialized methods and metadata determined from the inputs.

I'm just proposing the idea with this draft, looking forward to getting some feedback.


oesteban:
wdyt @effigies @mgxd @mattcieslak ?

oesteban:
I'm not sure this is really necessary. Will reopen if the final solution on top of #114 is based on this PR (see https://github.com/oesteban/sdcflows/compare/enh/fieldmap-objects...oesteban:rf/issues-20-21-26)

Issue #101, opened 2020-04-21
RFC: New heuristics table


oesteban:
In the context of nipreps/dmriprep#97, I'm trying to design a more generalizable set of heuristics when deciding the SDC estimation strategy. This is also related to nipreps/dmriprep#70.

This is likely to replace the ``fieldmap_wrangler`` eventually: https://github.com/nipreps/sdcflows/blob/02a824e3aeb6c0bd69c9643afc4c1d735d18cea8/sdcflows/workflows/base.py#L258-L294

The implementation will probably take the shape of a workflow, with an ``IdentityInterface`` for an ``outputnode`` where fields are the names of the EPI targets, for later selection.

Each of these fields will contain a list of fielmaps (which will be written out to the derivatives folder following the proposal in #26, with the adaptation of https://github.com/nipreps/sdcflows/issues/26#issuecomment-608034049), in the order and estimated as prescribed in the following table:

This heuristic table is to be applied **per participant**.
  - In the case of DWI, a target can be any ``dwi`` dataset under ``dwi/``.
  - In the case of BOLD, a target will be first any ``sbref`` and then any ``bold`` in the absence of ``sbref`` with ``IntendedFor``.

| Priority | Data                                                                                                              | Behavior                                                                                                                                                                                             |
| -------- | ----------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 0        | 2+ EPIs with different PE but common ``IntendedFor`` under ``fmap/``                                              | i) Estimate fieldmap (pepolar); ii) Apply fieldmap to ``IntendedFor`` targets                                                                                                                        |
| 1        | 1+ EPIs with common PE and ``IntendedFor`` under ``fmap/`` with 1+ target in ``IntendedFor`` with different PE.   | i) Estimate fieldmap (pepolar), using ``IntendedFor`` targets as alternative PEs; ii) Apply fieldmap to ``IntendedFor`` targets                                                                      |
| 2        | 2+ EPIs with different PE but no ``IntendedFor`` under ``fmap/``                                                  | Expand ``IntendedFor`` to all targets and apply 0?                                                                                                                                                   |
| 3        | 1+ EPIs with common PE and no ``IntendedFor`` under ``fmap/`` with 1+ target with different PE.                   | Expand ``IntendedFor`` to all targets and apply 1?                                                                                                                                                   |
[...]

Issue #20, opened 2019-07-09
SDC: Revise heuristics and method priority


oesteban:
In #1246, @satra proposed the following priority (I added those brackets):

>   1. Explicit B0 fieldmaps
>   2. [PE-polar] SE EPIs with multiple directions
>   3. [PE-polar] BOLD runs with multiple directions
>   4. [PE-polar?] BOLD runs without multiple directions

At this moment, priorities are:

  1. [PE-polar] SE EPIs with **opposed** directions (i.e., using RL and AP not possible together).
  2. Explicit B0 fieldmaps
  3. GRE/Phase difference fieldmaps
  4. Fieldmap-less (SyN) if ``--use-sdc-syn``.

In sum, we are missing the following options:

1. phase1/2 difference fieldmap
2. PE-polar: One polarity SE EPI + complementary polarity SBRef, BOLD or lowb (for diffusion).
3. (maybe) I'm still processing what option 4 meant - see #1246.

This issue attempts to provide a clear set of heuristics and priorities for SDC alternatives. We should also consider whether there is interest in allowing orthogonal PE blips EPIs (in contrast to current implementation that only allows opposed PE blips). If so, then we would need to move away from `3dQwarp` and pull `topup` back.

Paging @satra, @effigies,  @mattcieslak

satra:
@mattcieslak and @oesteban - the use case @mattcieslak brings up is a common one. we have collected b0s like that, where the diffusion series cannot even be used to fit a DTI model (i.e. doesn't have 6 directions). according to BIDS, there is no good place to put this image. we put it into the dwi directory for now, but it cannot be processed by any model.

oesteban:
> there is no good place to put this image. we put it into the dwi directory for now, but it cannot be processed by any model.

Is this still the situation?

We should write a PR and (IMHO) this should go under `anat/` (after all, the b0 is just a T2* contrast for anatomical reference - when no dwis are present).

satra:
@oesteban - in our case it cannot be used for diffusion modeling in some cases, but in other cases it can.  but i don't know what the current rationale is for what goes into anat. if `b0` goes into `anat`, could `sbref` go into `anat`? i think the semantics of these are still closer to their original intent, i.e. diffusion and func.

oesteban:
Sending this back to *fMRIPrep* proper, for the same reason I just closed nipreps/sdcflows#101 .

PR #114, opened 2020-10-23
ENH: New objects for better representation of fieldmap estimation

oesteban:
This PR attempts to provide a more reliable framework to build fieldmap estimation workflows.
Implicitly, it will help addressing issues regarding data conformity (e.g., #63, #64, #65) and also ease larger refactors such as ~#20~ [nipreps/fmriprep#2328], #21, #26, and #101.

@mattcieslak I think that on top of this PR we could consider some of the nitty-gritty details that you have previously encountered as problematic with QSIPREP. I'm referring, for instance, to checking shimming information if available. WDYT?


sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+
+    find_meta = attr.ib(kw_only=True, default=True, type=bool, repr=False)
+    """Enable/disable automated search for corresponding metadata."""
+
+    @path.validator
+    def check_path(self, attribute, value):
+        """Validate a fieldmap path."""
+        if isinstance(value, BIDSFile):
+            value = Path(value.path)
+        if isinstance(value, str):
```
mgxd:
```suggestion
        elif isinstance(value, str):
```

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+
+        if not value.is_file():
+            raise FileNotFoundError(
+                f"File path <{value}> does not exist, is a broken link, or it is not a file"
+            )
+
+        if not str(value).endswith((".nii", ".nii.gz")):
+            raise ValueError(f"File path <{value}> does not look like a NIfTI file.")
+
+        suffix = str(value).rpartition(".nii")[0].rpartition("_")[-1]
```
mgxd: will require adding an import, but I think this would be cleaner as a regex
```suggestion
        suffix = re.search('(?<=_)\w+(?=.nii)', str(value)).group()
```

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+            if not (
+                set(("TotalReadoutTime", "EffectiveEchoSpacing")).intersection(
+                    self.metadata.keys()
+                )
+            ):
+                raise MetadataError(
+                    f"Missing readout timing information for <{self.path}>."
+                )
+
+        if self.suffix == "fieldmap" and "Units" not in self.metadata:
```
mgxd: no need to hit anymore lines
```suggestion
        elif self.suffix == "fieldmap" and "Units" not in self.metadata:
```

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+            raise MetadataError(f"Missing 'Units' for <{self.path}>.")
+
+        if self.suffix == "phasediff" and (
```
mgxd:
```suggestion
        elif self.suffix == "phasediff" and (
```

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+            "EchoTime1" not in self.metadata or "EchoTime2" not in self.metadata
+        ):
+            raise MetadataError(
+                f"Missing 'EchoTime1' and/or 'EchoTime2' for <{self.path}>."
+            )
+
+        if self.suffix in ("phase1", "phase2") and ("EchoTime" not in self.metadata):
```
mgxd:
```suggestion
        elif self.suffix in ("phase1", "phase2") and ("EchoTime" not in self.metadata):
```

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+
+    def __attrs_post_init__(self):
+        """Validate metadata and additional checks."""
+        self.entities = parse_file_entities(str(self.path))
+        self.suffix = self.entities.pop("suffix")
+        extension = self.entities.pop("extension").lstrip(".")
+
+        # Automatically fill metadata in when possible
+        # TODO: implement BIDS hierarchy of metadata
+        if self.find_meta:
```
mgxd, line 190: this is probably fine, but what if the `find_meta` attribute was replaced with `bids_layout`, allowing the user to pass in an already initialized layout? then you could leverage methods like `find_metadata()`

sdcflows/fieldmaps.py
```diff
@@ -0,0 +1,345 @@
...
+            raise MetadataError(
+                f"Missing 'EchoTime1' and/or 'EchoTime2' for <{self.path}>."
+            )
+
+        if self.suffix in ("phase1", "phase2") and ("EchoTime" not in self.metadata):
+            raise MetadataError(f"Missing 'EchoTime' for <{self.path}>.")
+
+
+@attr.s(slots=True)
+class FieldmapEstimation:
```
mgxd, line 232: is there a reason for this to be a class, rather than a function returning sources / FM estimation method?

mgxd (commented):
gave this a quick first pass but overall like the idea, reminds me of the spaces refactor within fmriprep. left a few suggestions / questions

sdcflows/fieldmaps.py
oesteban: I'm always wary of regexes, if we're to use them then we need to test them

sdcflows/fieldmaps.py
oesteban: I wanted to assume that a BIDS layout might not be initialized. There are some other assumptions for the dataset to be BIDS-like (e.g., this really needs the input to have a suffix).

But the idea here is that we just need to insert a Nipype rename node to generate a name with the right suffix and then it should work regardless the file is in a BIDS structure or not (which may happen often, as well).

sdcflows/fieldmaps.py
oesteban: The attribute `method` keeps a state of the estimation (i.e. either unknown or a type of estimation). Also, I think a class can achieve similar things to a function, while having the responsibilities more clear and readable.

mgxd: Usually when I see a class with the only method being initialization, I think it can be stripped down to a function (`estimate_fieldmap_method()`) and achieve the same amount of clarity / readability, with the benefit of reducing some bloat. Would `method` ever be altered after this has been initialized?

sdcflows/fieldmaps.py
mgxd: I tested with all current modality keys with unix / windows paths and it extracted fine - any other tests you can think of?
and just to polish this up a little
```suggestion
        suffix = re.search('(?<=_)\w+(?=.nii)', value.name).group()
```

sdcflows/fieldmaps.py
oesteban: The function you suggest already exists - https://github.com/nipreps/sdcflows/blob/fef4ac623910b08d1f95d47d397d1c19dad0e45c/sdcflows/workflows/base.py#L259-L295

It is all but pretty or reliable. And yet, it doesn't implement most of the validation checkpoints in this PR or the metadata. Then the staging of the actual estimation happens on a different function - https://github.com/nipreps/sdcflows/blob/fef4ac623910b08d1f95d47d397d1c19dad0e45c/sdcflows/workflows/base.py#L22-L256 that has other validation points and decisions.

I wanted to start these objects because I was uncomfortable with the existing code. That code was also not very convenient for standalone implementations (#8) and while it seems to work acceptably for fMRIPrep (acceptable because, for instance, we currently replicate the processing of the same fieldmap for all runs in a dataset, if all are pointed by just one `IntendedFor` (i.e., one fieldmap). However, when I'm trying to use this on dMRIPrep, well it just does not work.

Objects are definitely easier to test, easier to document (sphinx does a fairly good job collecting docstrings) and more maintainable in the long run because they only have one, simple objective.

One thing I'm still internally debating is whether `FieldmapEstimation` should have a `get_workflow` or such, that directly returns the nipype workflow object built and with the inputs and metadata set (i.e., ready to connect the output to some derivatives sink workflow and correction workflows). I'm leaning towards doing it. Any feedback on this will be very very much appreciated.

sdcflows/fieldmaps.py
oesteban: The fact that I had to modify a little your suggestion does not help my tranquility :D : https://github.com/nipreps/sdcflows/blob/f551841b92605b5475e898b385c2774afcb41490/sdcflows/fieldmaps.py#L175

e.g.,
```Python
>>> re.search('(?<=_)\w+(?=.nii)', "sub-01_acq-nii_T1w.nii.gz").group()
'acq'
```

Issue #71, opened 2019-11-29
Cleaner edges when unwrapping with just extracted magnitude (no mask)


kimsin98:
I do not understand the reason yet, but fieldmap edges are consistently cleaner when PRELUDE is ran without a mask.

The reason could be because PRELUDE expects either
1. Original magnitude + mask
2. Brain extracted magnitude

but the current workflow provides both.

As of now, the unwrapping sometimes has too much noise on edges that the edge cleaning postprocessing is not enough.

AKSoo:
I checked PRELUDE's source code, and the reason seems to be that without mask, it generates a mask that further filters out some outlier values (below p2 + 0.1(p98 - p2) where p2 and p98 are 2nd/98th percentiles).

Unless the field of view is insanely tight, I don't think applying that threshold would cause any harm, and edges always look cleaner with it.

kimsin98:
I checked PRELUDE's code. Without mask, it generates a mask by filtering outliers values below `p2 + 0.1 * (p98 - p2)` where p2, p98 are 2nd, 98th percentiles (as described in [original article](https://doi.org/10.1002/mrm.10354)). On brain extracted magnitude images, this results in slightly tighter mask. There is no hard theory behind this threshold, but as the authors describe, it seems to give good results with PRELUDE in general.

I don't think applying this threshold would hurt unless the field of view is insanely tight.

Of course, a more sophisticated method for generating the mask could make this irrelevant. But until #23 is addressed, this should help.

oesteban:
I think this is not an issue anymore after #119 - I consistently see that edges do not bias the B-Spline estimation and noise in them is correctly removed. Please reopen if there is some other benefit to letting PRELUDE calculate the mask based on percentiles, but I'd say that in the context of #119 you would prefer to keep a more liberal mask.

Issue #22, opened 2019-07-10
Denoise kernel size


mattcieslak:
I'm finding for some older phase1/phase2 fieldmaps that it's necessary to set the kernel size to 5 for this node: https://github.com/poldracklab/sdcflows/blob/0.0.2/sdcflows/workflows/phdiff.py#L98

Is there a downside to setting the default to 5? Or  should we make this a parameter?

effigies:
Does it make sense to set a target smoothness instead of a straight kernel smoothing, as with [3dBlurToFWHM](https://afni.nimh.nih.gov/pub/dist/doc/program_help/3dBlurToFWHM.html)? Then we're less dependent on the quality of the input data.

oesteban:
Hopefully, this will not be necessary after #14, but it doesn't hurt to first get this done.

mattcieslak:
For now I have the kernel size only go up to 5 if there are two phase maps (#30) and the size stays 3 for phasediff images. #30 also re-organizes the workflows so that #14 can be used on these fieldmaps

oesteban:
I've set 5 as the default for now, and I'm pretty happy with the results.

I'm also with @effigies that `3dBlurToFWHM` is a very cheap but effective solution to the problem.

Finally, we need to investigate the resampling when moving the fieldmap to the target EPI dataset because I'm under the impression that it is introducing some little spikes in it.

mattcieslak:
Isn't `3dBlurToFWHM` doing Gaussian smoothing instead of a median filter? I thought the main purpose was to remove spikes, which would end up getting spread out spatially if you just smooth with a Gaussian kernel.

oesteban:
Isolated spikes would be removed with a Gaussian filter easily - median filters are better for salt&pepper kind of noise. It doesn't hurt to try.

But yes, you are right - the median filter is just to remove spikes

oesteban:
Superseded by #119.

PR #119, opened 2020-11-14
ENH: A 3D tensor B-Spline approximator and extrapolator

### LINKED ISSUE #14: Improve/finish Phasediff SDC regularization with B-Splines

### LINKED ISSUE #72: Check FUGUE extrapolation
As @AKSoo pointed out in https://github.com/poldracklab/fmriprep/pull/1886#issuecomment-559634776, this extrapolation should happen only along the PE-axis.

This issue is a reminder to check that end, until we have a better solution (#14).

### LINKED ISSUE #105: > Question-- do the reports need to visualize the inhomogeneity outside of the brain or do you think we should just mask it so that users don't get alarmed by the amount of out-of-brain field inhomogeneity?
> Question-- do the reports need to visualize the inhomogeneity outside of the brain or do you think we should just mask it so that users don't get alarmed by the amount of out-of-brain field inhomogeneity?

This question is THE QUESTION. On the one hand, I do agree it is alarming (especially to the non-familiar eye). On the other hand, I think it is nice to compare how different methods regularize outside the brain differently. For instance, the traditional FUGUE correction (the infamous `epiunwarp.fsl` script) extrapolates (constant) the values at brain edges along the PE axis. That produces [this kind of "rays"](https://neurostars.org/t/fmriprep-fieldmap/6000):
Screen Shot 2020-02-06 at 2 17 52 PM (1)

Would you mind filing an issue about this at [SDCflows](https://github.com/nipreps/sdcflows/issues/new)?

_Originally posted by @oesteban in https://github.com/nipreps/dmriprep/pull/97#issuecomment-640927659_

oesteban:
This PR finally adds an implementation for B-Spline smoothing and extrapolation of fieldmaps. It also makes the outputs of all estimation methods more consistent, as the outputs of TOPUP are two - the field and the B-Spline coefficients supporting the field.

Improving over TOPUP, this B-Spline modeling allows for multi-level B-Splines. Typically, you will want to use two grids of B-Splines. One with few control points to capture the global, low-spatial-frequency bias that makes the outline of the brain align better with the anatomy after correction, and with more dense coverage of the FoV to capture the more local high-frequency warping typically found at temporal lobes and the ventromedial prefrontal cortex.

~~This base implementation does not include a B-Spline interpolator interface that is necessary if you move the control points to the target EPI you want to undistort (which is the better way of moving the field, rather than the dense field representation). To make this possible, we will need to "un-project" the coefficients field and bring them into alignment with the original fieldmap when these are not "plumb" (i.e., scanner coordinate system is not aligned with the data array axes).~~ (EDITED 11/25 - it does include the interpolator now)

References: #71, #22.
Resolves: #72.
Resolves: #14.

---
EDITED - Text below was included on 11/25/2020, when the PR was nearing a conclusion.

## CHANGES
* Add the ``BSplineApprox`` interface: this is the nuts-and-bolts of this PR. This interface implements an approximation method to fit a grid of tensor-product cubic B-Spline basis, regularized with ridge regression. The output of this interface are coefficients maps (one for each resolution level of the B-Spline basis) embedded in NIfTI files. These NIfTI files have an affine aligned with the original fieldmap, but with the voxel sizes of the selected B-Spline basis.
* Add the ``Coefficients2Warp`` interface, which takes in coefficients NIfTI files (like those generated by ``BSplineApprox``, and when a conversion interface is set in place, those generated by TOPUP) and interpolates/extrapolates the field. It also converts from Hz to mm. To do so, the full affine of the target EPI being corrected is applied - meaning that it will perform correction along the PE axis considering that the affine can be "oblique" (see https://github.com/nipreps/sdcflows/pull/119/files#diff-c4cb2b592bcb50f08366cd3726e2ea4ff006fe1e43db8ff4b2e20cc6dd14161dR282). In other words, it does not naively multiply by the pixel spacing assuming that the image is "plumb" (using AFNI's nomenclature).
  **BUGFIX** - While working on this interface, an important bug calculating the Total Readout Time from the ``EffectiveEchoSpacing`` metadata has been solved - if the echo spacing info is "effective" it means that parallel acquisition and other acceleration factors have been already accounted for in the given figure (see https://github.com/nipreps/sdcflows/pull/119/files#diff-f9d322bceffe55f3a5c8b80d0c03c1a923ee838531b19ca12ad2d9435ceb4df9R107-R115).
* [Workflows] Adapted ``init_fmap_wf`` so that now, all types of fieldmaps (phasediff, phase1/2, and fieldmap) are smoothed with ``BSplineApprox``. That has also allowed dropping the old niflow and a bunch of "artisanal" settings from the good-old, mitical `epiunwarp.fsl` script of MGH. Now, fieldmaps are more consistent with TOPUP outputs, as both options have the same relevant outputs:
    * a fieldmap reference (magnitude file / EPI file)
    * a preprocessed fieldmap (in Hz), given in alignment with the fieldmap reference
    * coefficients file(s), in Hz, given in alignment with the preprocessed fieldmap (for the case of ``BSplineApprox``, only) - this is where we need to work on TOPUP's coefficients file, to store the BSpline grid separation and rotation w.r.t. canonical on the NIfTI's header.
* [Workflows] Revised the ``init_coeff2epi_wf`` workflow so that coefficients are moved into the target EPI space without resampling them. To do so, the affine of the NIfTI file is rotated and recentered following ANTs registration outputs (see https://github.com/nipreps/sdcflows/pull/119/files#diff-7121558e9d85fc90428b6d6b2f87d9586e7ff8682789782d847a831c12e80b1eR130). IMHO this is one of the coolest tiny details of this new implementation that will make it killer altogether. One test runs an SBRef-to-magnitude registration, see, e.g., this report: https://1436-189236569-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_space-sbref_fieldmap.svg
* [Workflows] Implemented a draft of ``init_unwarp_wf`` where only the coefficients moved (without resampling, remember) on to the target's space are used to reconstruct the fieldmap. This is a real breakthrough w.r.t. existing implementation of fieldmaps, esp. non-PEPOLAR:
    * The fieldmap resampled using the projected coefficients looks pretty good (see https://1436-189236569-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_task-rest_dir-LR_desc-fieldmap_bold.svg). For reference, the original coefficients file was generated by another test - see https://1436-189236569-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_desc-phasediff_fieldmap.svg - the apparent differences in intensity are caused by the more liberal mask of the SBRef, compared to the magnitude.
    * The correction does look good: https://1436-189236569-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_space-sbref_fieldmap.svg - *don't get distracted by those funny artifacts outside the brain*, that's because the extrapolated field there is very strong and signal is being picked up from far away. However that should be masked out. The important issue here is having the brain match the anatomical image, and that is clearly achieved with a pretty low-resolution BSpline grid. I would expect better results when running in *production* where we can apply a two-levels BSplines pyramid.
* Tests - as codecov registers, the diff is 89% covered (which is pretty nice). I'm sure some additional tests could be added, but I think for the time being this is already great.
* Documentation - revised existing documentation and improved some formulas. Attempted to make these formulas more descriptive of our implementation. Added lots of documentation about the B-Splines implementation.

NB. AFAIK, the implementation of B-Spline in numpy/scipy fall short for what we needed here. For the 2D case I could have reused very specific spline evaluation functions but nothing exists for the 3D case that we could have leveraged.


oesteban:
Okay, this is finally working. Please check the reports after this smoothing:

* Phasediff 1: [before (master)](https://697-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-1/figures/sub-1_acq-v2_desc-phasediff_fieldmap.svg) and [after (this PR)](https://731-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-1/figures/sub-1_acq-v2_desc-phasediff_fieldmap.svg)
* Phasediff 2: [before (master)](https://697-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-1/figures/sub-1_acq-v4_desc-phasediff_fieldmap.svg) and [after (this PR)](https://731-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-1/figures/sub-1_acq-v4_desc-phasediff_fieldmap.svg)
* Phasediff HCP: [before (master)](https://697-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_desc-phasediff_fieldmap.svg) and [after (this PR)](https://731-189292208-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-HCP101006/figures/sub-HCP101006_desc-phasediff_fieldmap.svg)

I think the results are pretty cool. Please note that, in _debug mode_ the fieldmap is not extrapolated outside the brain mask, and that the B-Spline grid is single and with the control points pretty far apart (that means the generated field will filter out high-frequency components of the original data).

There's perhaps one missing check to do - there might be a scaling component because the field is centered about zero, it seems that the dips and peaks could be fitted a bit more closely.

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+        elif self.inputs.recenter == "mean":
+            data -= np.mean(data[mask])
+
+        # Calculate spatial location of voxels, and normalize per B-Spline grid
+        mask_indices = np.argwhere(mask)
+        fmap_points = (
+            oriented_nii.affine.astype("float32") @ (
+                np.vstack((mask_indices.T, np.ones((1, mask_indices.shape[0],), dtype=int)))
+            )
+        )[:3].T
```
effigies: Is this just [`apply_affine`](https://github.com/nipy/nibabel/blob/591a1bd9cfb1d39b6834b5cf432fcbc8a76e3a7b/nibabel/affines.py#L17-L84)?

```suggestion
        fmap_points = nb.affines.apply_affine(oriented_nii.affine.astype("float32"), mask_indices)
```

Might need to add some `.T`s in there...

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+
+        # Store outputs
+        out_name = str(
+            Path(
+                fname_presuffix(
+                    self.inputs.in_data, suffix="_field", newpath=runtime.cwd
+                )
+            ).absolute()
+        )
```
effigies: Is there a situation where `runtime.cwd` isn't absolute? If so, we should fix that in nipype, rather than hack around it.

```suggestion
        out_name = fname_presuffix(self.inputs.in_data, suffix="_field", newpath=runtime.cwd)
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # Fit the model
+        model = lm.Ridge(alpha=self.inputs.ridge_alpha, fit_intercept=False)
+        model.fit(regressors.T, data[mask])
+
+        interp_data = np.zeros_like(data)
+        interp_data[mask] = (
+            (np.array(model.coef_) @ regressors)  # Interpolation
+            .astype("float32")
+        )
```
effigies: `interp_data` is float32.

```suggestion
        interp_data[mask] = (np.array(model.coef_) @ regressors  # Interpolation
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if not self.inputs.extrapolate:
+            return runtime
+
+        bg_indices = np.argwhere(~mask)
+        bg_points = (
+            oriented_nii.affine.astype("float32") @ (
+                np.vstack((bg_indices.T, np.ones((1, bg_indices.shape[0],), dtype=int)))
+            )
+        )[:3].T
```
effigies:
```suggestion
        bg_points = nb.affines.apply_affine(oriented_nii.affine.astype("float32"), bg_indices)
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+
+        extrapolators = np.vstack(
+            [bspline_weights(bg_points, level) for level in bs_levels]
+        )
+        interp_data[~mask] = (
+            (np.array(model.coef_) @ extrapolators)  # Extrapolation
+            .astype("float32")
+        )
```
effigies:
```suggestion
        interp_data[~mask] = (np.array(model.coef_) @ extrapolators  # Extrapolation
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+    ctl_spacings = [float(sp) for sp in level.header.get_zooms()[:3]]
+    ncoeff = level.dataobj.size
+    ctl_points = (
+        level.affine.astype("float32") @ (
+            np.vstack((
+                np.argwhere(np.asanyarray(level.dataobj) == 0.0).astype("float32").T,
+                np.ones((1, ncoeff), dtype="float32")
+            ))
+        )
+    )[:3].T
```
effigies: This can be another apply_affines case.

Also, are you guaranteed to have exactly zeros? Do you want to use `isclose`?

```suggestion
    ctl_points = nb.affines.apply_affine(
        level.affine.astype("float32"),
        np.argwhere(np.isclose(level.dataobj, 0)))
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,249 @@
...
+    # Center both images
+    im_center = img.affine @ np.hstack((0.5 * (im_shape - 1), 1))
+    bs_center = bs_affine @ np.hstack((0.5 * (bs_shape - 1), 1))
+    bs_affine[:3, 3] = im_center[:3] - bs_center[:3]
+
+    return nb.Nifti1Image(np.zeros(bs_shape, dtype="float32"), bs_affine)
+
+
+def bspline_weights(points, level):
+    """Calculate the tensor-product cubic B-Spline weights for a list of 3D points."""
```
effigies: I think future maintainers will thank you for a reference for the formulas.

effigies (commented):
I haven't worked through the workflow, but here are some comments on the interfaces. Please ping me again if I don't finish this tomorrow.

sdcflows/interfaces/bspline.py
oesteban: Should be zero because the image object is initialized with zeros before, but yes, this should be safer.

oesteban:
Rebased on top of #123 - https://github.com/oesteban/sdcflows/compare/enh/new-unwarp-api...oesteban:enh/bspline-regularization

oesteban:
@mgxd after the latest force-push to remove a little update of `docs/conf.py` that I have moved directly into master, I think this is ready for you to have a look. I have edited the PR's message to include some information that I think will help you navigate through this monster.

Comments from @effigies and @mattcieslak (or anyone else) are also welcome of course (although I may defer to the moment we merge `dev/1.4.0` into `master` for the feedback if I feel that merging this will allow me to get further, and I know Chris is not available these days, for a start).

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+        desc="spacing between B-Spline control points",
+    )
+    ridge_alpha = traits.Float(
+        0.01, usedefault=True, desc="controls the regularization"
+    )
+    recenter = traits.Enum(
+        "mode",
+        "median",
+        "mean",
+        "no",
```
mgxd: IMO a little cleaner
```suggestion
        False,
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+        interp_data = np.zeros_like(data)
+        interp_data[mask] = np.array(model.coef_) @ regressors  # Interpolation
+
+        # Store outputs
+        out_name = fname_presuffix(
+            self.inputs.in_data, suffix="_field", newpath=runtime.cwd
+        )
+        hdr = fmapnii.header.copy()
+        hdr.set_data_dtype("float32")
+        nb.Nifti1Image(interp_data, fmapnii.affine, hdr).to_filename(out_name)
```
mgxd: should allow nifti2 as well?
```suggestion
        fmapnii.__class__(interp_data, fmapnii.affine, hdr).to_filename(out_name)
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+        self._results["out_field"] = out_name
+
+        index = 0
+        self._results["out_coeff"] = []
+        for i, (n, bsl) in enumerate(zip(ncoeff, bs_levels)):
+            out_level = out_name.replace("_field.", f"_coeff{i:03}.")
+            nb.Nifti1Image(
```
mgxd: ?
```suggestion
            bsl.__class__(
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+                ),
+                bsl.affine,
+                bsl.header,
+            ).to_filename(out_level)
+            index += n
+            self._results["out_coeff"].append(out_level)
+
+        # Write out fitting-error map
+        self._results["out_error"] = out_name.replace("_field.", "_error.")
+        nb.Nifti1Image(
```
mgxd:
```suggestion
        fmapnii.__class__(
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+
+        bg_indices = np.argwhere(~mask)
+        bg_points = apply_affine(fmapnii.affine.astype("float32"), bg_indices)
+
+        extrapolators = np.vstack(
+            [bspline_weights(bg_points, level) for level in bs_levels]
+        )
+        interp_data[~mask] = np.array(model.coef_) @ extrapolators  # Extrapolation
+        self._results["out_extrapolated"] = out_name.replace("_field.", "_extra.")
+        nb.Nifti1Image(interp_data, fmapnii.affine, hdr).to_filename(
```
mgxd:
```suggestion
        fmapnii.__class__(interp_data, fmapnii.affine, hdr).to_filename(
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+class _Coefficients2WarpInputSpec(BaseInterfaceInputSpec):
+    in_target = File(exist=True, mandatory=True, desc="input EPI data to be corrected")
+    in_coeff = InputMultiObject(
+        File(exists=True),
+        mandatory=True,
+        desc="input coefficients, after alignment to the EPI data",
+    )
+    ro_time = traits.Float(1.0, usedefault=True, desc="EPI readout time (s).")
```
mgxd: i'm a bit concerned with setting this by default

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+
+        data = np.zeros(targetnii.shape, dtype="float32")
+        data[allmask == 1] = np.squeeze(np.vstack(coeffs).T) @ np.vstack(weights)
+
+        hdr = targetnii.header.copy()
+        hdr.set_data_dtype("float32")
+        self._results["out_field"] = fname_presuffix(
+            self.inputs.in_target, suffix="_field", newpath=runtime.cwd
+        )
+        nb.Nifti1Image(data, targetnii.affine, hdr).to_filename(
```
mgxd:
```suggestion
        targetnii.__class__(data, targetnii.affine, hdr).to_filename(
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+        self._results["out_warp"] = fname_presuffix(
+            self.inputs.in_target, suffix="_xfm", newpath=runtime.cwd
+        )
+        # Compose a vector field
+        field = np.zeros((data.size, 3), dtype="float32")
+        field[..., phaseEncDim] = data.reshape(-1)
+        aff = targetnii.affine.copy()
+        aff[:3, 3] = 0.0
+        field = nb.affines.apply_affine(aff, field).reshape(fieldshape)
+        warpnii = nb.Nifti1Image(
```
mgxd:
```suggestion
        warpnii = targetnii.__class__(
```

sdcflows/interfaces/bspline.py
```diff
@@ -0,0 +1,387 @@
...
+    # Calculate the shape of the B-Spline grid
+    im_extent = im_zooms * (im_shape - 1)
+    bs_shape = (im_extent // bs_zooms + 3).astype(int)
+
+    # Center both images
+    bs_affine[:3, 3] = apply_affine(img.affine, 0.5 * (im_shape - 1)) - apply_affine(
+        bs_affine, 0.5 * (bs_shape - 1)
+    )
+
+    return nb.Nifti1Image(np.zeros(bs_shape, dtype="float32"), bs_affine)
```
mgxd:
```suggestion
    return img.__class__(np.zeros(bs_shape, dtype="float32"), bs_affine)
```

sdcflows/interfaces/reportlets.py
```diff
@@ -77,6 +77,9 @@ def _generate_report(self):
         fmapdata = fmapnii.get_fdata()
         vmax = max(abs(np.percentile(fmapdata[maskdata], 99.8)),
                    abs(np.percentile(fmapdata[maskdata], 0.2)))
+        if self.inputs.apply_mask:
+            fmapdata[~maskdata] = 0
+            fmapnii = nb.Nifti1Image(fmapdata, fmapnii.affine, fmapnii.header)
```
mgxd:
```suggestion
            fmapnii = fmapdata.__class__(fmapdata, fmapnii.affine, fmapnii.header)
```

mgxd (commented):
Some comments / suggestions from my first pass. Overall I think this looks great.

sdcflows/interfaces/bspline.py
oesteban:
```suggestion
    ro_time = traits.Float(mandatory=True, desc="EPI readout time (s).")
```

sdcflows/interfaces/tests/test_bspline.py
```diff
@@ -0,0 +1,57 @@
...
+        in_coeff=str(tmp_path / "coeffs.nii.gz"),
+        pe_dir="j-",
+    ).run()
+
+    # Approximate the interpolated target
+    test2 = BSplineApprox(
+        in_data=test1.outputs.out_field,
+        in_mask=str(tmp_path / "target.nii.gz"),
+        bs_spacing=[(4, 6, 8)],
+        recenter="no",
```
oesteban:
```suggestion
        recenter=False,
```

sdcflows/interfaces/tests/test_bspline.py
```diff
@@ -0,0 +1,57 @@
...
+    coeff = (np.random.random(size=gridnii.shape) - 0.5) * 500
+    coeffnii = nb.Nifti1Image(coeff.astype("float32"), gridnii.affine, gridnii.header)
+    coeffnii.to_filename(tmp_path / "coeffs.nii.gz")
+
+    os.chdir(tmp_path)
+    # Check that we can interpolate the coefficients on a target
+    test1 = Coefficients2Warp(
+        in_target=str(tmp_path / "target.nii.gz"),
+        in_coeff=str(tmp_path / "coeffs.nii.gz"),
+        pe_dir="j-",
```
oesteban, line 44:
```suggestion
        pe_dir="j-",
        ro_time=1.0,
```

Issue #129, opened 2020-11-26
Attempting to instantiate two ``FieldmapEstimation`` from the same sources should fail


oesteban:
Meaning that we probably need a dictionary of FieldmapEstimation objects hashed by the sources set, and return the existing object (or fail?) when attempting to create a new one.

PR #130, opened 2020-11-26
ENH: Keep a registry of already-used identifiers (and auto-generate new)

### LINKED ISSUE #125: Implement `B0FieldIdentifier` in `FieldmapEstimation` object
To facilitate the implementation of bids-standard/bids-specification#622 if that comes through, with the following:

1. Extend `FieldmapEstimation` with an `id` property, and to keep a register of unique IDs (to ensure there're no duplicities).
1. Read the `id` from the BIDS metadata. If the `id` is not present, generate one automatically and forward it to the derivatives.
1. Make sure the `id` makes it to the subworkflow name to ensure correct Nipype performance and avoid collisions.

oesteban:
This is the projection of the ``B0FieldIdentifier`` metadata proposed for the BIDS spec. Having an identifier for each estimation strategy is the most robust way of minimizing the computational load and reliably keep track of fieldmap information. 

Followed-up-by: #129.
Resolves: #125.

### DIFF
--- a/sdcflows/fieldmaps.py
+++ b/sdcflows/fieldmaps.py
@@ -9,6 +9,9 @@ from bids.utils import listify
 from niworkflows.utils.bids import relative_to_root
 
 
+_unique_ids = set()
+
+
 class MetadataError(ValueError):
     """A better name for a specific value error."""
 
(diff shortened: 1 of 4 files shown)


oesteban:
Okay, I think this PR would be the most interesting part of the [big refactor](https://github.com/nipreps/sdcflows/milestone/2?closed=1) I'm undertaking on [`dev/1.4.0`](https://github.com/nipreps/sdcflows/tree/dev/1.4.0).

It also starts with the implementation of the `B0FieldIdentifier` metadata, if it finally makes it into the spec (if not, there are technical reasons to still keep this - although we cannot count on PyBIDS to do the job in that case).

I think this tracking helps justify the utilization of the `FieldmapEstimation` object (instead of just a function), which was a concern for @mgxd. The idea is to keep building on the `FieldmapEstimation` class and make it the _factory_ for the creation of the estimation workflows (i.e., it will be the recommended API to `sdcflows.workflows.fit`.

Thoughts?

oesteban:
Okay, I'm going to continue with the overhaul - sorry for not allowing the adequate review window to properly eyeball through all the changes.

PR #144, opened 2020-12-04
Massive merge of the new 2.0 API

oesteban:
I don't expect a lot of feedback given the dates and the massive size of this PR, but any comments will be welcome. I think we can better debug these changes by exercising the new API with fMRIPrep and dMRIPrep, so my next steps will go towards that direction (meaning, I will not wait a lot before merging this in). I assume the risks of rushing this merge, but a new API is very badly needed - so let's work from this.

EDIT

## I'm preparing everything to ship the new API with a bump of the major release number - hence we'll be starting the 2.x series.


oesteban:
Okay, going for it. I really want to see this moving along.

PR #150, opened 2020-12-11
ENH: Write ``fmapid-`` entity in Derivatives

### LINKED ISSUE #146: Write B0FieldmapIdentifier as an entity in Derivatives
This will allow disentangling the different fieldmaps estimated.

oesteban:
Derived from ``B0FieldIdentifier``, to allow easier matching of ``B0FieldSource``.

Depends: nipreps/niworkflows#585.
Resolves: #146.

### DIFF
--- a/sdcflows/workflows/base.py
+++ b/sdcflows/workflows/base.py
@@ -71,17 +71,19 @@ def init_fmap_preproc_wf(
 
     workflow = Workflow(name=name)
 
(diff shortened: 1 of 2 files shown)


oesteban:
Working locally, I don't see a reason why this would not work on Circle (and if that happened, I can always go back and fix master to help the sprint advance).

Issue #169, opened 2020-12-29
BUG: Optimization of coefficient calculation has broken the unwarp workflow


oesteban:
Since merging #157 with 39728a86816e014796b4df4171ac2b83748810e1, the reportlet showing correction (using topup for estimation) is broken.

PR #170, opened 2020-12-29
FIX: Misconfigured test of unwarping workflow

oesteban:
The test seemed broken after a PR was merged (see #169), but that PR actually improved the implementation.

Here, the coefficients file is rewritten with a newly generated file from TOPUP and after applying `TOPUPCoeffReorient`.


oesteban:
Okay, test looking better now - https://2316-189236569-gh.circle-artifacts.com/0/tmp/tests/sdcflows/sub-101006/figures/sub-101006_task-rest_dir-LR_desc-corrected_bold.svg

PR #174, opened 2021-01-20
FIX: Fast & accurate brain extraction of magnitude images without FSL BET

### LINKED ISSUE #23: Too liberal mask of the magnitude image

oesteban:
Using mathematical morphology and the random walker algorithm (as it is implemented in scikit-image).


oesteban:
This also affects #67, not necessarily resolving it (but, in combination with the B-Spline regularization, ds000217 --or any other limited FoV dataset-- should be fine)

PR #178, opened 2021-01-26
enh: add override to default bspline distance

eilidhmacnicol:
The addition of the EPI brain masking workflow is not quite compatible with rodents as seen here:
bspline_distance-default

The proposed PR provides an optional integer to the workflow to override the default b-spline fitting distance. 

This was used to change the b-spline fitting distance to 15, from the default 300:
bspline_distance-15


oesteban (approved):
LGTM

PR #190, opened 2021-03-23
RF: Divide ambiguous debug parameter into smaller, more focused parameters

mgxd:
This PR aims to separate the sometimes ambiguous "debug" argument into two clearer options:

- `debug` - when set to True, produces intermediate files / forces re-runs. The main usage of this flag is to track down a specific step where the implementation is failing.
- `sloppy` - when set to True, run operations with faster, but less precise, values.

So far, this is just altering workflows that are called directly within `FieldmapEstimation.get_workflow()`


oesteban (approved):
LGTM

PR #191, opened 2021-04-05
ENH: Add estimation method description to outputs

mgxd:
Adds a `method` output field to estimation workflows, to allow easy inspection of type of correction applied.

- Descriptions are ported from the `1.3` series of SDCFlows


mgxd:
let's get this in - we can consider moving to inspecting the metadata once/if? https://github.com/bids-standard/bids-specification/pull/622 is in

Issue #194, opened 2021-04-22
Memory issues in ``Coefficients2Warp``


oesteban:
When using higher resolutions of the B-Spline field, this interface has serious problems to fit all it needs into memory.

PR #195, opened 2021-04-22
FIX: Address memory issues of ``Coefficients2Warp``

oesteban:
This PR ends with the dual implementation of the evaluation of B-Spline weights on the grid of the approximated map. On the one hand, we had a ``grid_bspline_weights`` function, that used vector operations, halfway optimized. On the other, I later wrote ``bspline_weights`` as the former was untenable by the approximation interface.

I vaguely remember having plans to unify them but left the code there untouched.

These changes should ease the memory issues of ``Coefficients2Warp`` (still hooked with the first implementation) - see #194.


oesteban:
This has been superseded by #203

PR #196, opened 2021-04-27
FIX: Convert TOPUP coefficients from mm to Hz

oesteban:


sdcflows/interfaces/bspline.py
```diff
@@ -592,7 +606,7 @@ def _move_coeff(in_coeff, fmap_ref, transform):
     return out
 
 
-def _fix_topup_fieldcoeff(in_coeff, fmap_ref, out_file=None):
+def _fix_topup_fieldcoeff(in_coeff, fmap_ref, pe_dir, ro_time=1.0, out_file=None):
```
mgxd, line 609: should there be a default for readout time?

sdcflows/interfaces/bspline.py
```diff
@@ -621,5 +635,10 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, out_file=None):
     coeffnii.header.set_qform(coeffnii.header.get_qform(coded=False), code=0)
     coeffnii.header.set_sform(newaff, code=1)
 
+    pe_axis = "ijk".index(pe_dir[0])
+    pe_sign = [1.0, -1.0][len(pe_dir) != 2]
```
mgxd: is this flipped? in any case, I think this can be a little more explicit
```suggestion
    pe_sign = -1.0 if pe_dir.endswith('-') else 1.0
```

mgxd: Just to confirm - my suggestion changed the behavior (i.e., `j- == -1.0`)

oesteban: Yes, I believe you are right.

oesteban:
This has been superseded by #203.

PR #199, opened 2021-05-04
FIX: Revise generation of the displacements field from coefficients

### LINKED ISSUE #198: Incorrect conversion of PE displacements to displacements field
Revise ``Coefficients2Warp`` because, in the case the PE axis is flipped w.r.t. canonical direction, the direction of the displacements is flipped twice.

However, the idea of projecting the distortion if the data are not "plumb" (i.e., their affine is oblique) is good.

oesteban:
Resolves: #198.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -284,7 +284,7 @@ class Coefficients2Warp(SimpleInterface):
         if obliquity(aff).max() * 180 / pi > 0.01:
             dirmat = np.eye(4)
             dirmat[:3, :3] = aff[:3, :3] / (voxel_sizes(aff) * io_orientation(aff)[:, 1])
-            field = nb.affines.apply_affine(aff, field)
+            field = nb.affines.apply_affine(dirmat, field)
 
         warpnii = targetnii.__class__(
             field.reshape(fieldshape)[:, :, :, np.newaxis, :], targetnii.affine, None
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -281,7 +281,7 @@ class Coefficients2Warp(SimpleInterface):
 
         # If coordinate system is oblique, project displacements through directions matrix
         aff = targetnii.affine
-        if (obliquity(aff).min() * 180 / pi) > 0.01:
+        if obliquity(aff).max() * 180 / pi > 0.01:
             dirmat = np.eye(4)
             dirmat[:3, :3] = aff[:3, :3] / (voxel_sizes(aff) * io_orientation(aff)[:, 1])
             field = nb.affines.apply_affine(aff, field)
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -2,9 +2,11 @@
 # vi: set ft=python sts=4 ts=4 sw=4 et:
 """Filtering of :math:`B_0` field mappings with B-Splines."""
 from pathlib import Path
+from math import pi
 import numpy as np
 import nibabel as nb
-from nibabel.affines import apply_affine
+from nibabel.affines import apply_affine, obliquity, voxel_sizes
+from nibabel.orientations import io_orientation
 
 from nipype import logging
 from nipype.utils.filemanip import fname_presuffix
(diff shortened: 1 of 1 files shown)


sdcflows/interfaces/bspline.py
```diff
@@ -264,24 +266,28 @@ def _run_interface(self, runtime):
...
-        field[..., phaseEncDim] = data.reshape(-1)
-        aff = targetnii.affine.copy()
-        aff[:3, 3] = 0.0
-        # Multiplying by the affine implicitly applies the voxel size to the shift map
-        field = nb.affines.apply_affine(aff, field).reshape(fieldshape)
+        field[..., pe_axis] = data.reshape(-1)
+
+        # If coordinate system is oblique, project displacements through directions matrix
+        aff = targetnii.affine
+        if (obliquity(aff).min() * 180 / pi) > 0.01:
```
effigies: Do you want `min` here instead of `max`? `min` implies that it only needs to be plumb on one axis, not all.

sdcflows/interfaces/bspline.py
```diff
@@ -264,24 +266,28 @@ def _run_interface(self, runtime):
...
+            dirmat = np.eye(4)
+            dirmat[:3, :3] = aff[:3, :3] / (voxel_sizes(aff) * io_orientation(aff)[:, 1])
+            field = nb.affines.apply_affine(aff, field)
```
effigies: I'm confused. `dirmat` isn't used. Should this be:

```suggestion
            dirmat = np.eye(4)
            dirmat[:3, :3] = aff[:3, :3] / (voxel_sizes(aff) * io_orientation(aff)[:, 1])
            field = nb.affines.apply_affine(dirmat, field)
```

I'm not quite following `dirmat`. We're getting the cosines, divided by the zooms, so it should be a unit rotation matrix, and then flipping R/L, A/P, S/I. Is this flip going to undo the direction encoded in `j`/`j-`?

oesteban: This is exactly the problem we are fixing. If the PE direction is `j-` and the affine has `RPI` orientation, then this will drop the flip of the two last axes in this case.

Similarly, if the PE direction is `j-` and the orientation is `PLS`, the directions matrix will be made `ARS`.

In other words, the nonzero component of the fieldmap needs to be projected through the oblique matrix so that the displacements are actually done along the PE axis by tools like `antsApplyTransforms` that work on world coordinates.

Did I manage to explain this?

sdcflows/interfaces/bspline.py
oesteban: Good catch, thanks

sdcflows/interfaces/bspline.py
oesteban: And yes, not using `dirmat` was an oversight. The idea is that dirmat gives you the unit vectors of the rotated axes, right-hand coordinates (i.e., no flips).

effigies:
Okay, I think I got it.

Suppose we have a plumb dataset in LPS with PE dir j- (A). Because we do nothing `dirmat` we can consider the identity.

Now suppose that we're 1deg rotated clockwise around the vertical axis (the LP plane is rotated around S). So moving positively in the j direction moves mostly posterior and very slightly left, and negatively moves anterior and slightly right. This is identical to rotating the RA plane around S. Rotating the whole field by a 1deg rotation must not include flips.

So I've convinced myself of that. I don't have time right now to think through whether we're applying the correct rotation.

oesteban:
> So moving positively in the j direction moves mostly posterior and very slightly left, and negatively moves anterior and slightly right.

Correct... on voxel coordinates.

The problem is that the oblique affine generates a "world" coordinate system, and there's no concept of "axes-flip" there. 

I'm not 100% sure of this, but I believe that an image oriented LAS would not pass the obliquity check (i.e., will be considered "plumb"). In this case, and if the PE direction is i-, the consideration for the positive/negative direction of the displacement is done already some lines above (and we don't double-flip).

If you want a tool that works in world coordinates to apply a transform along a certain data array axis, and the axes are not aligned with the world coordinates system, you indeed need to project the displacement on to the axes. But you don't change the direction.

effigies:
Okay, I've talked myself into why we rotate in the same direction as the world and not backwards, so this seems right. My final point of unclarity is: Why is the distortion in world space and not in voxel space? If I'm oriented at 45deg to the pedir, shouldn't I still smear along the pedir and not to my anterior?

oesteban:
The grid in which the fieldmap is sampled is indeed in voxel coordinates. Also the displacements field that ANTs takes in to apply the transform but the displacements themselves are in world coordinates

effigies:
So should we be applying the inverse rotation?

oesteban:
Ugh that's sensible. I need to think it through.

PR #200, opened 2021-05-05
ENH: Double-check conversion from TOPUP to standardized fieldmaps

oesteban:
This PR extends the TOPUP workflow with a few nodes to apply the correction outside TOPUP/ApplyTOPUP.

This extension allows the developer double-check that the conversion from TOPUP outpus is done correctly, and the correction looks good.

Depends: #203.


oesteban:
This code made me sure that we are applying the fieldmap correctly on NiBabies, @mgxd

oesteban:
I'm going for it and for cutting a release on a Friday 🚀

Issue #201, opened 2021-05-05
Identify, implement and test the different interactions of PE and EPI's x-forms


oesteban:
First identified in debugging nipreps/nibabies#55, SDCFlows might not be generating correct displacements fields out of fieldmaps (``Coefficients2Warp`` interface) when there are swapped axis and/or flipped directions.

Heavily related to the PR and corresponding conversation taking place in #199.

oesteban:
(duplicated)

PR #204, opened 2021-05-06
ENH: :math:`B_0` fieldmap unwarping object

### LINKED ISSUE #194: Memory issues in ``Coefficients2Warp``

### LINKED ISSUE #203: A transform class based on NiTransforms for fieldmaps
Write a transform class specific for fieldmap-informed unwarping.

Members:

- A coefficients field.
- The reference EPI (distorted EPI grid): shape, affine

Inputs of the `resample()` member (besides those of  the transform base class of NiTransforms):
-  total readout time
- The phase encoding direction code of the reference EPI

Implementation of `resample()`:

- Move the coefficients field on to the reference (currently implemented within the bspline interfaces module)
- Calculate the bspline weights (exposing the sparse matrix of weights will help shave downstream calculations off)
- Interpolate the fieldmaps on the target grid
- Convert the fieldmap from Hz to voxels (i.e., multiply by total readout time in secs and apply the polarity +/-)
- Calculate the target voxel coordinates (this is the neat feature of this object - we don't need to convert into mm, and hence fight with oblique affines, as opposed to using antsApplyTransforms).

Other reasons for the existence of this object:
- Easier to implement targetted unit tests
- Integration with other projects (e.g., I would immediately use this in https://github.com/nipreps/eddymotion to implement on-the-fly SDC of each diffusion MRI volume prior head motion estimation, which is pretty cool, btw).
- Encapsulation of B-Spline interpolation -related routines, which currently are under a `bspline` module for nipype interfaces.

Does this make sense @mgxd @effigies ?

@tsalo could be very interested in this too, as it sideways relates to his idea for dynamic fieldmaps.

oesteban:
First implementation of the unwarping object.
Example utilization:

```Python
import nibabel as nb
from sdcflows import transform as tf

coeffs = nb.load("sub-01_dir-AP_epi_idx-000_merged_base_fieldcoef_fixed.nii.gz")
xfm = tf.B0FieldTransform(coeffs=coeffs)
unwarped = xfm.apply("sub-01_dir-AP_epi_idx-000.nii.gz", pe_dir="j-", ro_time=0.0597409)
unwarped.to_filename("epi_unwarped.nii.gz")
```

Gridded evaluation of the tensor-product B-Spline weights is done using sparse matrices and fixes the memory issues.
Minimizes the problems of #201, as now the transform is applied in voxel coordinates without need for tinkering with the affine matrix.

Resolves: #194.
Resolves: #203.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -8,6 +8,7 @@ Information on specific functions, classes, and methods.
 
    api/sdcflows.fieldmaps
    api/sdcflows.interfaces
+   api/sdcflows.transform
    api/sdcflows.utils
    api/sdcflows.viz
    api/sdcflows.workflows
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -133,15 +133,15 @@ class BSplineApprox(SimpleInterface):
                 if weights is None
                 else sparse_vstack((weights, gbsw(fmapnii, level)))
             )
-        weights = weights.toarray()
-        regressors = weights[:, mask.reshape(-1)]
+
+        regressors = weights.T.tocsr()[mask.reshape(-1), :]
 
         # Fit the model
         model = lm.Ridge(alpha=self.inputs.ridge_alpha, fit_intercept=False)
-        model.fit(regressors.T, data[mask])
+        model.fit(regressors, data[mask])
 
         interp_data = np.zeros_like(data)
-        interp_data[mask] = np.array(model.coef_) @ regressors  # Interpolation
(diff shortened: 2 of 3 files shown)


oesteban:
This may be of interest to @arokem too.

oesteban:
Okay, this is pretty much it. I'll merge as it is going to unlock myself in both NiBabies and dMRIPrep.

It would be nice to create more unit tests more targetted, and to have an idea of how well the memory/speed issues have been eliminated.

PR #217, opened 2021-07-07
ENH: Add slice padding to TOPUP

mgxd:
Because of the configuration we're using, TOPUP fails when given an image with an odd number of slices. This adds a new interface `PadSlice` that adds an additional empty slice to images with odd numbered slices.

~~Once I test this, I'll convert out of draft mode.~~

A few questions:
- Does the pad interface need to take into account image orientation?
- Do we want to remove the added slice after calculating TOPUP correction?

### DIFF
--- a/sdcflows/interfaces/utils.py
+++ b/sdcflows/interfaces/utils.py
@@ -269,6 +269,11 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
 
 class _PadSlicesInputSpec(BaseInterfaceInputSpec):
     in_file = File(exists=True, mandatory=True, desc="3D or 4D NIfTI image")
+    axis = traits.Int(
+        2,
+        usedefault=True,
+        desc="The axis through which slices are stacked in the input data"
+    )
 
 
 class _PadSlicesOutputSpec(TraitedSpec):
@@ -277,12 +282,17 @@ class _PadSlicesOutputSpec(TraitedSpec):
 
 
 class PadSlices(SimpleInterface):
+    """
+    Check an image for uneven slices, and add an empty slice if necessary
+
+    This intends to avoid TOPUP's segfault without changing the standard configuration
+    """
     input_spec = _PadSlicesInputSpec
     output_spec = _PadSlicesOutputSpec
 
     def _run_interface(self, runtime):
         self._results["out_file"], self._results["padded"] = _pad_num_slices(
-            self.inputs.in_file, runtime.cwd,
+            self.inputs.in_file, self.inputs.axis, runtime.cwd,
         )
         return runtime
 
@@ -409,7 +419,7 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
 def _pad_num_slices(in_file, ax=2, newpath=None):
     """
     Ensure the image has even number of slices to avert TOPUP's segfault.
-    
+
     Check if image has an even number of slices.
     If it does, return the image unaltered.
     Otherwise, return the image with an empty slice added.
--- a/sdcflows/interfaces/utils.py
+++ b/sdcflows/interfaces/utils.py
@@ -267,6 +267,26 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
     _copy_header_map = {"output_image": "input_image"}
 
 
+class PadSlicesInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc="3D or 4D NIfTI image")
+
+
+class PadSlicesOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc="The output file with even number of slices")
+    padded = traits.Bool(desc="Indicator if the input image was padded")
+
+
+class PadSlices(SimpleInterface):
+    input_spec = PadSlicesInputSpec
+    output_spec = PadSlicesOutputSpec
+
+    def _run_interface(self, runtime):
+        self._results["out_file"], self._results["padded"] = _pad_num_slices(
+            self.inputs.in_file, runtime.cwd,
+        )
+        return runtime
+
+
 def _flatten(inlist, max_trs=50, out_dir=None):
     """
     Split the input EPIs and generate a flattened list with corresponding metadata.
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
         masknii.__class__(masknii.dataobj, epinii.affine, hdr).to_filename(out_files[2])
 
     return out_files
+
+
+def _pad_num_slices(in_file, newpath=None):
+    """
+    Check if image has an even number of slices.
+    If it does, return the image unaltered.
+    Otherwise, return the image with an empty slice added.
+
+    Parameters
+    ----------
+    img : :obj:`str` or :py:class:`~nibabel.spatialimages.SpatialImage`
+        3D or 4D NIfTI image
+
+    Returns
+    -------
+    file : :obj:`str`
+        The output file with even number of slices
+    padded : :obj:`bool`
+        Indicator if the input image was padded.
+
+    """
+    import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+    import numpy as np
+
+    img = nb.load(in_file)
+    if img.shape[2] % 2 == 0:
+        return in_file, False
+
+    pwidth = [(0,0)] * len(img.shape)
+    pwidth[2] = (0, 1)
+    padded = np.pad(img.dataobj, pwidth)
+    hdr = img.header
+    hdr.set_data_shape(padded.shape)
+    out_file = fname_presuffix(in_file, suffix="_padded", newpath=newpath)
+    img.__class__(padded, img.affine, header=hdr).to_filename(out_file)
+    return out_file, True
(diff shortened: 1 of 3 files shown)


mgxd:
@oesteban I'm a bit confused on how the best way to handle this added slice further down the pipeline.

TOPUP now proceeds without error, but https://github.com/nipreps/sdcflows/blob/a3765456544b27ed653911b601a14603529beca5/sdcflows/workflows/fit/pepolar.py#L152-L154
complains of a shape mismatch between the regridded reference fieldmap and the coefficients file.

Would it be better to:
- Assign the padded image as the reference? or
- Remove a slice from the coeffs generated by TOPUP?

oesteban:
> complains of a shape mismatch between the regridded reference fieldmap and the coefficients file.

where exactly? I would expect this, but I don't know where precisely. My first intuition would be to use the padded image as reference for the coefficients.

mgxd:
This is the check

https://github.com/nipreps/sdcflows/blob/d2023f59a278a57cc7e15888fd545c349b8d82d2/sdcflows/interfaces/bspline.py#L447-L454

For now, I'm testing padding the reference image as well.

oesteban:
Yes, use the padded reference for this interface too 👍

sdcflows/interfaces/utils.py
```diff
@@ -267,6 +267,26 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
     _copy_header_map = {"output_image": "input_image"}
 
 
+class PadSlicesInputSpec(BaseInterfaceInputSpec):
```
oesteban:
```suggestion
class _PadSlicesInputSpec(BaseInterfaceInputSpec):
```

sdcflows/interfaces/utils.py
```diff
@@ -267,6 +267,26 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
...
+    in_file = File(exists=True, mandatory=True, desc="3D or 4D NIfTI image")
+
+
+class PadSlicesOutputSpec(TraitedSpec):
```
oesteban:
```suggestion
class _PadSlicesOutputSpec(TraitedSpec):
```

sdcflows/interfaces/utils.py
```diff
@@ -267,6 +267,26 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
...
+    out_file = File(exists=True, desc="The output file with even number of slices")
+    padded = traits.Bool(desc="Indicator if the input image was padded")
+
+
+class PadSlices(SimpleInterface):
+    input_spec = PadSlicesInputSpec
+    output_spec = PadSlicesOutputSpec
```
oesteban:
```suggestion
    input_spec = _PadSlicesInputSpec
    output_spec = _PadSlicesOutputSpec
```

In niworkflows, and here, we use the underscore so then sphinx renders the interface (using nipype's extension of sphinx) more nicely.

sdcflows/interfaces/utils.py
```diff
@@ -267,6 +267,26 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
...
+class PadSlicesInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc="3D or 4D NIfTI image")
+
+
+class PadSlicesOutputSpec(TraitedSpec):
+    out_file = File(exists=True, desc="The output file with even number of slices")
+    padded = traits.Bool(desc="Indicator if the input image was padded")
+
+
+class PadSlices(SimpleInterface):
```
oesteban, line 284: Please add a comment describing the new interface

sdcflows/interfaces/utils.py
```diff
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
         masknii.__class__(masknii.dataobj, epinii.affine, hdr).to_filename(out_files[2])
 
     return out_files
+
+
+def _pad_num_slices(in_file, newpath=None):
+    """
+    Check if image has an even number of slices.
```
oesteban, line 423:
```suggestion
    Ensure the image has even number of slices to avert TOPUP's segfault.
    
    Check if image has an even number of slices.
```

There should be an empty line between the oneliner description of the function and the long description.

sdcflows/interfaces/utils.py
```diff
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
...
+    padded : :obj:`bool`
+        Indicator if the input image was padded.
+
+    """
+    import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+    import numpy as np
+
+    img = nb.load(in_file)
+    if img.shape[2] % 2 == 0:
```
oesteban: With rodents, we may want to allow another axis. May we use add a default argument `ax=2` to the signature instead?
```suggestion
    if img.shape[ax] % 2 == 0:
```

sdcflows/interfaces/utils.py
```diff
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
         masknii.__class__(masknii.dataobj, epinii.affine, hdr).to_filename(out_files[2])
 
     return out_files
+
+
+def _pad_num_slices(in_file, newpath=None):
```
oesteban:
```suggestion
def _pad_num_slices(in_file, ax=2, newpath=None):
```

sdcflows/interfaces/utils.py
```diff
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
...
+    """
+    Check if image has an even number of slices.
+    If it does, return the image unaltered.
+    Otherwise, return the image with an empty slice added.
+
+    Parameters
+    ----------
+    img : :obj:`str` or :py:class:`~nibabel.spatialimages.SpatialImage`
+        3D or 4D NIfTI image
+
```
oesteban, line 433:
```suggestion
    ax : :obj:`int`
        The axis through which slices are stacked in the input data.

```

sdcflows/interfaces/utils.py
```diff
@@ -384,3 +404,40 @@ def _reoblique(in_epi, in_plumb, in_field, in_mask=None, newpath=None):
...
+    import nibabel as nb
+    from nipype.utils.filemanip import fname_presuffix
+    import numpy as np
+
+    img = nb.load(in_file)
+    if img.shape[2] % 2 == 0:
+        return in_file, False
+
+    pwidth = [(0,0)] * len(img.shape)
+    pwidth[2] = (0, 1)
```
oesteban:
```suggestion
    pwidth[ax] = (0, 1)
```

oesteban (approved):
Left a few comments.

mgxd:
I think this is ready, I've opened up #220 to note the unrelated circle failure

effigies:
I would want to use a few orientations to test that the same ras+ locations have the same values, to make sure we're correctly padding with respect to the affine.

mgxd:
forgot to follow up here, sorry for the delay - @effigies can you take a look at fdb9dbc and lmk what's needed to push this along?

sdcflows/interfaces/utils.py
```diff
@@ -267,6 +267,36 @@ class DenoiseImage(_DenoiseImageBase, _CopyHeaderInterface):
     _copy_header_map = {"output_image": "input_image"}
 
 
+class _PadSlicesInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc="3D or 4D NIfTI image")
+    axis = traits.Int(
```
effigies, line 272: We never set this. Do we need to check `SliceEncodingDirection`?

julfou81:
Hi, to address this issue with odd number of slices with popup, what is recommended now by FSL is to use a specific configuration file that will not use subsampling and will not be affected by a dimension being an odd number. Of note this will increase the execution time of topup in that case.

https://www.jiscmail.ac.uk/cgi-bin/webadmin?A2=FSL;6c4c9591.2002

Would that be an option in SDCflows  to use different configuration files for topup depending on the matrix size of the images?

oesteban:
Thanks for the reference - it is pretty good. I don't think, though, the rationale behind such a recommendation applies here:

1. We are not removing one slice, but adding instead.
2. There is an extremely strong assumption in Jesper's message that the grids of the template and the target EPI are the same - which is the case when you tailor the protocol to use FSL TOPUP, but from a theoretical point of view might be very problematic.
3. Those configuration files basically create denser B-Spline basis to account for the design problem of not allowing control points to fall off-grid. Which is IMHO pretty sloppy, and worse, makes results of TOPUP across different matrix sizes uncomparable. I feel it is pretty poor advice. The separation of knots of the B-Spline field has a physical motivation, assuming they are tuned w.r.t. the expected smoothness of the field along each axis.

julfou81:
Thank you @oesteban for your quick and detail comments. Very interesting. I did not think about those aspects. I will have to think it through.
 On my side I followed Jesper advice for some diffusion data where I had some issue using the slice2vol correction in FSL-eddy when I was changing the number of slices by adding or removing one slice on dataset with odd number of slices.

oesteban:
Adding a slice should not affect if appended to the right (as in, correct) end of the axis.

julfou81:
ok, sounds good! I will see if I manage to make this work with eddy and eddy_quad doing it right with the slice padding. At the time I made a quick test that did not work and then found Jesper's answer that was pretty convenient for me and did not bother more about changing the number of slice on preprocessing on datasets with odd number of slices. Thank you again for your comments!

Issue #220, opened 2021-08-05
CI: `cache_test_data` job failing


mgxd:
Step: Setup git-annex
```
#!/bin/bash -eo pipefail
if [[ ! -d /usr/lib/git-annex.linux ]]; then
  wget -O- http://neuro.debian.net/lists/focal.us-ca.full | tee /etc/apt/sources.list.d/neurodebian.sources.list
  apt-key adv --recv-keys --keyserver hkp://pool.sks-keyservers.net:80 0xA5D32F012649A5A9
  apt update && apt-get install -y --no-install-recommends git-annex-standalone
fi
git config --global user.name 'NiPreps Bot'
git config --global user.email 'nipreps@gmail.com'

--2021-08-05 13:30:35--  http://neuro.debian.net/lists/focal.us-ca.full
Resolving neuro.debian.net (neuro.debian.net)... 129.170.233.11
Connecting to neuro.debian.net (neuro.debian.net)|129.170.233.11|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 260
Saving to: ‘STDOUT’

-                   100%[===================>]     260  --.-KB/s    in 0s      
deb http://neurodeb.pirsquared.org data main contrib non-free
#deb-src http://neurodeb.pirsquared.org data main contrib non-free
deb http://neurodeb.pirsquared.org focal main contrib non-free
#deb-src http://neurodeb.pirsquared.org focal main contrib non-free

2021-08-05 13:30:35 (48.4 MB/s) - written to stdout [260/260]

Executing: /tmp/apt-key-gpghome.nZbM820qat/gpg.1.sh --recv-keys --keyserver hkp://pool.sks-keyservers.net:80 0xA5D32F012649A5A9
gpg: keyserver receive failed: No name

Exited with code exit status 2
```

We probably just need to point to another keyserver

effigies:
The sks pool is supposed to be redundant. Should look at what's going on over there.

effigies:
Ah: https://unix.stackexchange.com/questions/656205/sks-keyservers-gone-what-to-use-instead

> This service is deprecated. This means it is no longer maintained, and new HKPS certificates will not be issued. Service reliability should not be expected.
> 
> Update 2021-06-21: Due to even more GDPR takedown requests, the DNS records for the pool will no longer be provided at all.

effigies:
@mgxd @oesteban What's the oldest maintenance branch we still submit PRs to?

effigies:
[...]

PR #222, opened 2021-08-05
MNT: Use keys.openpgp.org over sks-keyservers

effigies:
Addresses #220 in `maint/1.3.x`.


effigies:
Okay, it looks like we need to backport more CircleCI fixes than were trivial.

mgxd:
@effigies can this be closed?

effigies:
We'll need to fix this if we ever need another sdcflows 1.3.x release. If not, we can close.

mgxd:
ah, this is tied to the LTS, probably a good idea to fix it up

PR #229, opened 2021-09-07
FIX: Update boilerplate ordering directives

mgxd:
Partially addresses https://github.com/nipreps/fmriprep/issues/2526


effigies (approved):
Along with some fixes that I will submit to fMRIPrep shortly, this works for me.

PR #232, opened 2021-09-10
ENH: Allow tools to specify a default PhaseEncodingDirection for SyN-SDC

effigies:
To recover SyN-SDC behavior in fMRIPrep, it's necessary to provide a default phase-encoding direction. Currently it would need to be added to the metadata, but it seems cleaner to make it an explicit argument.

### DIFF
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -26,7 +26,7 @@ from contextlib import suppress
 from pathlib import Path
 
 
-def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
+def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False, default_pedir=None):
     """
     Apply basic heuristics to automatically find available data for fieldmap estimation.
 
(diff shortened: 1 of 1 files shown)


effigies:
Should be considered alongside https://github.com/nipreps/fmriprep/pull/2530.

oesteban:
Sorry, I am really behind on these developments. As I commented on the other thread, this doesn't feel all right.

If the user does know the `PhaseEncodingDirection`, it should be found in the metadata.
If they know the PE axis, but they do not know the polarity, we should send a PR to the BIDS specs so that this can be written with the metadata, and/or perhaps allow nonstandard metadata for this particular in fMRIPrep/SDCFlows, but adding a shortcut like this seems the wrong move, as it has very very low friction for the user on something potentially dangerous, and it is very likely this is lost in reporting (i.e., likely to become not-transparent).

As opposed to the case of readout time (for which I propose that internally SDCFlows may use 1.0 if unknown, provided that the user is aware that the units of the fieldmap are unknown and it should be used will caution), in this case if you don't know the PE direction, then you'd rather not run fieldmapless at all.

Alternatively, we could run fieldmapless in BOTH possible axes and then pick the one that gave a better fit - although this goes into the rabbit hole of how you decide between two registration processes. But at least you can show both results with the visual report, and very clearly state that this was a decision made by fMRIPrep. Follow-up questions would include what happens if fMRIPrep determines different PE directions for two runs that were certainly acquired with the same direction (which is not unlikely to happen, I'd say).

effigies:
Fair enough.

PR #234, opened 2021-09-15
ENH: Improve support of 4D in ``sdcflows.interfaces.bspline.ApplyCoeffsField``

### LINKED ISSUE #233: Extension of ``sdcflows.interfaces.bspline.ApplyCoeffsField``
In order to fix nipreps/fmriprep#2210, several improvements of the machinery under this interface must be improved:

- Accept 4D files
- Accept a list of affines to pass in head-motion parameters
- Accept a transform file to project b-spline coefficients (instead of doing this elsewhere).

Future iterations will involve:
- Accept other transforms (t1 to epi, std to t1, etc) implementing all-in-one-shot resamplings

oesteban:
This PR revises the implementation of the core unwarping interface, by enabling `sdcflows.transforms.B0FieldTransform` to project the coefficients representing the fieldmap into each volume of the input through the head-motion correction matrix, thereby moving the fieldmap with the head motion of that frame and hence applying time-step-wise correction.

I'm not confident this will really add substantial accuracy, but theoretically is the way to go. Given the sophistication (and the pervasive interpolation for each frame of the bspline field), this is not implemented elsewhere (except, perhaps, for applyTOPUP).

Resolves: #233.
Initiated-by: nipreps/fmriprep#2210.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -42,7 +42,6 @@ from nipype.interfaces.base import (
 
 from sdcflows.transform import grid_bspline_weights as gbsw, B0FieldTransform
 
-
 LOW_MEM_BLOCK_SIZE = 1000
 DEFAULT_ZOOMS_MM = (40.0, 40.0, 20.0)  # For human adults (mid-frequency), in mm
 DEFAULT_LF_ZOOMS_MM = (100.0, 100.0, 40.0)  # For human adults (low-frequency), in mm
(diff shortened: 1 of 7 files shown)


oesteban:
Okay, this is starting to look pretty good. I'm testing the new code within fMRIPrep and the feeling is optimistic.

It is very cool that `B0FieldTransform` is capable of producing hmc-ed and sdc-ed time-series, with a single interpolation step. Extension to sampling on surfaces and/or having further transforms to pre-concatenate does not seem like daunting future iterations at this point.

I might need to add some little changes here and there as fMRIPrep continues, but overall this is ready for your judgment.

effigies (commented):
This overall looks sensible to me. Having a bit of trouble keeping track of which way the HMC transforms are going. I kind of expected more affine inversions, but I'm assuming that you're keeping proper track...

oesteban:
The following steps are:

i. Adding support to prepend additional transforms (e.g., epi-to-anatomical, anatomical-to-standard)
ii. Modulate intensity through the Jacobian of the SDC transform, prior to resampling
iii. An interface to project surfaces (requires (i))

As the three steps require the foundations set in this PR, I think the best path forward is merging this one. I'm aware I'm reducing coverage by ~4%, so I'll think about adding meaningful tests.

PR #237, opened 2021-09-30
FIX: Generation of RAS displacements fields from VSMs

### LINKED ISSUE #31: Identify, implement and test the different interactions of PE and EPI's x-forms
Comes from https://github.com/poldracklab/fmriprep/issues/1493#issuecomment-459851011

### LINKED ISSUE #218: Unexplained flip of the PE direction under some unknown circumstances
Using @mgxd's dataset in the debugging of the fMRIPrep + SDCFlows integration, we seem to have hit a difficult bug for which I haven't found the failure point (and hence the fix).

In order to permit progress on nipreps/fmriprep#2392, I'm separating this problem that seems to do with some idiosyncratic parameter (likely related to orientation headers) that makes the problem apparent.

### LINKED ISSUE #236: ``B0FieldTransform`` botches the displacements field with flipped axes
It remains unclear when this is an axis flip and when is just incorrect fields being generated, but either case, this requires attention.

Linked: #218.

oesteban:
After much thought, I have come to the conclusion that the VOX2RAS affine must be applied in all cases, regardless of whether the dataset is oblique or not.

After all, the VSM (voxel-shift-map) is no more than a regularly gridded field of voxel coordinates. The transformation between voxels and mm is biyective and univocal, and formalized by the affine matrix of the image.

This PR simplifies the calculation, and theoretically should resolve most of the issues we are experiencing when resampling data.

Resolves: #218.
Resolves: #236.

Related: nipreps/fmriprep#2210.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -229,12 +229,15 @@ class B0FieldTransform:
         ijk_deltas[..., pe_axis] = vsm.reshape(-1)
 
         # To convert from VSM to RAS field we just apply the affine
-        aff = self.shifts.affine
+        aff = self.shifts.affine.copy()
+        aff[:3, 3] = 0  # Translations MUST NOT be applied, though.
         xyz_deltas = nb.affines.apply_affine(aff, ijk_deltas)
         xyz_nii = nb.Nifti1Image(
             # WARNING: ITK NIfTI fields are 5D (have an empty 4th dim).
             #          Hence, the ``np.newaxis``.
-            xyz_deltas.reshape(fieldshape)[:, :, :, np.newaxis, :], aff, None
+            xyz_deltas.reshape(fieldshape)[:, :, :, np.newaxis, :],
+            self.shifts.affine,
+            None,
         )
         xyz_nii.header.set_intent("vector", (), "")
         xyz_nii.header.set_xyzt_units("mm")
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -215,38 +215,30 @@ class B0FieldTransform:
             A NIfTI 1.0 object containing the distortion.
 
         """
-        from math import pi
-        from nibabel.affines import voxel_sizes, obliquity
-        from nibabel.orientations import io_orientation
-
-        # Generate warp field
-        data = self.shifts.get_fdata(dtype="float32").copy()
+        # Set polarity & scale VSM (voxel-shift-map) by readout time
+        vsm = self.shifts.get_fdata().copy()
         pe_axis = "ijk".index(pe_dir[0])
-        pe_sign = -1.0 if pe_dir.endswith("-") else 1.0
-        pe_size = self.shifts.header.get_zooms()[pe_axis]
-        data *= pe_sign * ro_time * pe_size
+        vsm *= -1.0 if pe_dir.endswith("-") else 1.0
+        vsm *= ro_time
 
-        fieldshape = tuple(list(data.shape[:3]) + [3])
+        # Shape of displacements field
+        fieldshape = tuple(list(vsm.shape[:3]) + [3])
 
-        # Compose a vector field
-        field = np.zeros((data.size, 3), dtype="float32")
-        field[..., pe_axis] = data.reshape(-1)
+        # Convert VSM to voxel displacements
+        ijk_deltas = np.zeros((vsm.size, 3), dtype="float32")
+        ijk_deltas[..., pe_axis] = vsm.reshape(-1)
 
-        # If coordinate system is oblique, project displacements through directions matrix
+        # To convert from VSM to RAS field we just apply the affine
         aff = self.shifts.affine
-        if obliquity(aff).max() * 180 / pi > 0.01:
-            dirmat = np.eye(4)
-            dirmat[:3, :3] = aff[:3, :3] / (
-                voxel_sizes(aff) * io_orientation(aff)[:, 1]
-            )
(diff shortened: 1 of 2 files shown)


oesteban:
@mgxd this (in combination with the latest fMRIPrep integration) should fix things for your dataset - please let me know

oesteban:
I'm writing some unit tests, just to be sure this is what we need.

sdcflows/transform.py
```diff
@@ -215,38 +215,33 @@ def to_displacements(self, ro_time, pe_dir):
...
-            field.reshape(fieldshape)[:, :, :, np.newaxis, :], aff, None
+        vsm *= -1.0 if pe_dir.endswith("-") else 1.0
+        vsm *= ro_time
+
+        # Shape of displacements field
+        fieldshape = tuple(list(vsm.shape[:3]) + [3])
+
+        # Convert VSM to voxel displacements
+        ijk_deltas = np.zeros((vsm.size, 3), dtype="float32")
+        ijk_deltas[..., pe_axis] = vsm.reshape(-1)
```
effigies: 2D array. Ellipses just make this less clear.

```suggestion
        ijk_deltas[:, pe_axis] = vsm.reshape(-1)
```

sdcflows/transform.py
```diff
@@ -215,38 +215,33 @@ def to_displacements(self, ro_time, pe_dir):
...
-            )
-            field = nb.affines.apply_affine(dirmat, field)
-
-        warpnii = nb.Nifti1Image(
-            field.reshape(fieldshape)[:, :, :, np.newaxis, :], aff, None
+        vsm *= -1.0 if pe_dir.endswith("-") else 1.0
+        vsm *= ro_time
+
+        # Shape of displacements field
+        fieldshape = tuple(list(vsm.shape[:3]) + [3])
```
effigies: Given that we're adding a newaxis later, why not:

```suggestion
        # Note that ITK NIfTI fields are 5D (have an empty 4th dimension)
        fieldshape = tuple(list(vsm.shape[:3]) + [1, 3])
```

sdcflows/transform.py
```diff
@@ -215,38 +215,33 @@ def to_displacements(self, ro_time, pe_dir):
...
+        ijk_deltas[..., pe_axis] = vsm.reshape(-1)
+
+        # To convert from VSM to RAS field we just apply the affine
+        aff = self.shifts.affine.copy()
+        aff[:3, 3] = 0  # Translations MUST NOT be applied, though.
+        xyz_deltas = nb.affines.apply_affine(aff, ijk_deltas)
+        xyz_nii = nb.Nifti1Image(
+            # WARNING: ITK NIfTI fields are 5D (have an empty 4th dim).
+            #          Hence, the ``np.newaxis``.
+            xyz_deltas.reshape(fieldshape)[:, :, :, np.newaxis, :],
```
effigies: Then this is just:

```suggestion
            xyz_deltas.reshape(fieldshape),
```

effigies (approved):
This makes sense to me. Made some very small suggestions, but feel free to reject without argument.

sdcflows/transform.py
```diff
@@ -215,38 +215,36 @@ def to_displacements(self, ro_time, pe_dir):
...
+        # To convert from VSM to RAS field we just apply the affine
+        aff = self.shifts.affine.copy()
+        aff[:3, 3] = 0  # Translations MUST NOT be applied, though.
+        xyz_deltas = nb.affines.apply_affine(aff, ijk_deltas)
+        if itk_format:
+            xyz_deltas[..., (0, 1)] *= -1.0
+
+        xyz_nii = nb.Nifti1Image(
+            # WARNING: ITK NIfTI fields are 5D (have an empty 4th dim).
+            #          Hence, the ``np.newaxis``.
```
oesteban: This comment is not necessary anymore.
```suggestion
```

oesteban:
I'm writing a test with a numerical phantom. I have been able to compare the correction with our `B0FieldTransform` object (i.e., in voxel space) and with `antsApplyTransforms` using the displacements field our object generates. With RAS, canonical affine it works. Will configure the test to explore obliquity, flips and axis ordering.

sdcflows/transform.py
```diff
@@ -215,38 +215,35 @@ def to_displacements(self, ro_time, pe_dir):
...
+        # Convert VSM to voxel displacements
+        ijk_deltas = np.zeros((vsm.size, 3), dtype="float32")
+        ijk_deltas[:, pe_axis] = vsm.reshape(-1)
+
+        # To convert from VSM to RAS field we just apply the affine
+        aff = self.shifts.affine.copy()
+        aff[:3, 3] = 0  # Translations MUST NOT be applied, though.
+        xyz_deltas = nb.affines.apply_affine(aff, ijk_deltas)
+        if itk_format:
+            xyz_deltas[..., (0, 1)] *= -1.0
```
effigies, line 238: Oh god. What a nasty bug. I think this deserves a comment.

```suggestion
            # ITK displacement vectors are in LPS orientation
            xyz_deltas[..., (0, 1)] *= -1.0
```

effigies (approved):
Approving again. Very convincing test!

oesteban:
> Very convincing test!

Yup, it doesn't tell you that we have a good implementation of fieldmaps, but we are safe to say that, once you have the fieldmap our tool and ANTs will apply the transform the same. So it indeed resolves the unit test :)

Merging in, we can include @mgxd on his data later down the line.

PR #239, opened 2021-10-01
FIX: SDC-SyN ("fieldmap-less") overhaul

### LINKED ISSUE #85: syn-sdc largely suboptimal results
I mentioned in Neurostars that the syn-sdc we get are not good. As suggested (I think by @effigies) I am posting a couple of examples here for your review.

Please note that my area of research focuses on basal ganglia (caudate and putamen more), so we focus a lot on the ventricles going into place after the correction.

These are cases where we have fieldmaps and compare fmb-sdc with syn-sdc figures. Since SVG files cannot be uploaded in github, I have added some screen recordings in the attached powerpoint file.

[Syn-SDC_subptimal.pptx](https://github.com/poldracklab/fmriprep/files/4149307/Syn-SDC_subptimal.pptx)

We don't have a lot cases without fieldmaps, but looks like those without fieldmaps are a lost case, particularly from Philips scanners.

### LINKED ISSUE #187: T1w-SyN: Is contrast inversion really necessary? (at least for mutual-information scales)
I think I've heard @mattcieslak asking about this, please let me know if I'm wrong (whether you made the comment and/or the content of the comment).

The question is that, most of the linear and nonlinear registration steps of our SyN implementation use Mattes. With this cost function, the contrast inversion just does nothing - while it is one more massaging step where we can harm the reliability of the overall process.

Maybe we could limit the contrast inversion only for the final CC ``antsRegistration`` step.

For the case of BOLD, it would be doubtful we need to invert the contrast at all in some cases. Because we are picking the frames with the strongest T1 contrast across all runs in the session with share shimming settings, the inversion is probably not needed anymore (even for the CC metric).

Thoughts?

(I'm giving this very low priority, but nonetheless noting it down for future reference)

EDIT: I just found the comment from Matt - https://github.com/nipreps/sdcflows/issues/68#issuecomment-562805659. It was right under my nose. Apologies.

oesteban:
In the context of nipreps/fmriprep#2530, this PR takes a deep revision on the SDC-SyN implementation. In passing, the problems about suboptimal registration (#85) are also addressed.


oesteban:
Okay, I think the registration parameters of both initialization and SyN steps, and the B-Spline approximation, are pretty good for the one subject I'm testing on from ds005. We probably want to merge this and maybe refine parameters as we test this on other datasets.

sdcflows/tests/test_transform.py
```diff
@@ -134,3 +134,26 @@ def test_displacements_field(tmpdir, testdata_dir, outdir, pe_dir, rotation, fli
                 f"_y-{rotation[1] or 0}_z-{rotation[2] or 0}.svg"
             ),
         ).run()
+
+
+@pytest.mark.parametrize("pe_dir", ["j", "j-", "i", "i-", "k", "k-"])
+def test_conversions(tmpdir, testdata_dir, pe_dir):
+    """Check idempotency."""
```
effigies:
```suggestion
    """Check inverse functions."""
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    # Set polarity & scale VSM (voxel-shift-map) by readout time
+    vsm = fmap_nii.get_fdata().copy() * (-ro_time if pe_dir.endswith("-") else ro_time)
+
+    # Shape of displacements field
+    # Note that ITK NIfTI fields are 5D (have an empty 4th dimension)
+    fieldshape = tuple(list(vsm.shape[:3]) + [1, 3])
+
+    # Convert VSM to voxel displacements
+    ijk_deltas = np.zeros((vsm.size, 3), dtype="float32")
+    ijk_deltas[:, "ijk".index(pe_dir[0])] = vsm.reshape(-1)
```
effigies: This feels like one too many things to keep in my head; had to go back to the diff to double-check my understanding. Would prefer to spend a line to have a variable name to connect these.

```suggestion
    pe_axis = "ijk".index(pe_dir[0])
    ijk_deltas = np.zeros((vsm.size, 3), dtype="float32")
    ijk_deltas[:, pe_axis] = vsm.reshape(-1)
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    xyz_deltas = nb.affines.apply_affine(aff, ijk_deltas)
+    if itk_format:
+        # ITK displacement vectors are in LPS orientation
+        xyz_deltas[..., (0, 1)] *= -1.0
+
+    xyz_nii = nb.Nifti1Image(
+        xyz_deltas.reshape(fieldshape),
+        fmap_nii.affine,
+        None,
+    )
```
effigies: Can save a couple lines here, though:

```suggestion
    xyz_nii = nb.Nifti1Image(xyz_deltas.reshape(fieldshape), fmap_nii.affine)
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    xyz_nii.header.set_intent("vector", name="SDC")
+    xyz_nii.header.set_xyzt_units("mm")
+    return xyz_nii
+
+
+def disp_to_fmap(xyz_nii, ro_time, pe_dir, itk_format=True):
+    """
+    Convert a displacements field into a fieldmap in Hz.
+
+    This is the dual operation to the previous function.
```
```suggestion
    This is the inverse operation to the previous function.
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    <sdcflows.workflows.fit.fieldmap.html#mjx-eqn-eq%3Afieldmap-2>`__.
+
+    Parameters
+    ----------
+    fmap_nii : :obj:`os.pathlike`
+        Path to a voxel-shift-map (VSM) in NIfTI format
+    ro_time : :obj:`float`
+        The total readout time in seconds (only if ``vsm=False``).
+    pe_dir : :obj:`str`
+        The ``PhaseEncodingDirection`` metadata value (only if ``vsm=False``).
```
effigies: I think the `vsm=False` comments are out-of-place here.
```suggestion
        The total readout time in seconds
    pe_dir : :obj:`str`
        The ``PhaseEncodingDirection`` metadata value
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    This is the dual operation to the previous function.
+
+    Parameters
+    ----------
+    xyz_nii : :obj:`os.pathlike`
+        Path to a displacements field in NIfTI format.
+    ro_time : :obj:`float`
+        The total readout time in seconds (only if ``vsm=False``).
+    pe_dir : :obj:`str`
+        The ``PhaseEncodingDirection`` metadata value (only if ``vsm=False``).
```
```suggestion
        The total readout time in seconds.
    pe_dir : :obj:`str`
        The ``PhaseEncodingDirection`` metadata value.
```

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    )
+
+    ijk_nii = nb.Nifti1Image(
+        ijk_deltas.reshape(xyz_nii.shape[:3]),
+        xyz_nii.affine,
+        None,
+    )
+    ijk_nii.header.set_intent("estimate", name="Delta_B0 [Hz]")
+    ijk_nii.header.set_xyzt_units("mm")
+    return ijk_nii
```
effigies: This attempts to more closely mirror the inverse function:

```suggestion
    pe_axis = "ijk".index(pe_dir[0])
    vsm = ijk_deltas[:, pe_axis].reshape(xyz_nii.shape[:3])
    scale_factor = -ro_time if pe_dir.endswith("-") else ro_time

    fmap_nii = nb.Nifti1Image(vsm / scale_factor, xyz_nii.affine)
    fmap_nii.header.set_intent("estimate", name="Delta_B0 [Hz]")
    fmap_nii.header.set_xyzt_units("mm")
    return fmap_nii
```

effigies (approved):
I think it makes sense. If I understand correctly, we're moving from histogram-based matching to finding a single deformation that optimally registers both intensity and the intensity-derived Laplacian (~edges) images?

Some comments (mostly readability) about the inverse functions.

sdcflows/transform.py
```diff
@@ -215,36 +217,108 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+    spatialimage : :obj:`nibabel.nifti.Nifti1Image`
+        A NIfTI 1.0 object containing the distortion.
+
+    """
+    # Set polarity & scale VSM (voxel-shift-map) by readout time
+    vsm = fmap_nii.get_fdata().copy() * (-ro_time if pe_dir.endswith("-") else ro_time)
+
+    # Shape of displacements field
+    # Note that ITK NIfTI fields are 5D (have an empty 4th dimension)
+    fieldshape = tuple(list(vsm.shape[:3]) + [1, 3])
```
mgxd: To avoid excess type conversion
```suggestion
    fieldshape = vsm.shape[:3] + (1, 3)
```

sdcflows/workflows/fit/syn.py
```diff
@@ -198,29 +204,52 @@ def init_syn_sdc_wf(
 """
     inputnode = pe.Node(niu.IdentityInterface(INPUT_FIELDS), name="inputnode")
     outputnode = pe.Node(
-        niu.IdentityInterface(["fmap", "fmap_ref", "fmap_coeff", "fmap_mask", "method"]),
+        niu.IdentityInterface(
+            ["fmap", "fmap_ref", "fmap_coeff", "fmap_mask", "out_warp", "method"]
```
mgxd, line 240: can you add `out_warp` to the docstring outputs?

sdcflows/workflows/fit/syn.py
```diff
@@ -236,13 +265,13 @@ def init_syn_sdc_wf(
...
     )
 
-    deoblique = pe.Node(CopyXForm(fields=["epi_ref"]), name="deoblique")
-    reoblique = pe.Node(Reoblique(), name="reoblique")
-
     # Set a manageable size for the epi reference
     find_zooms = pe.Node(niu.Function(function=_adjust_zooms), name="find_zooms")
     zooms_epi = pe.Node(RegridToZooms(), name="zooms_epi")
 
+    # histmatch = pe.Node(niu.Function(function=match_histogram),
```
mgxd: can we just remove this?

mgxd (approved):
lgtm, minor comments

oesteban:
Thanks for the feedback :)

Two late additions:

- 43712de932f2dbd3625aacd49b7e5643b9e15530 where we make sure the ``restrict_deformation`` input of ANTs' registration is in RAS coordinates (instead of voxel coordinates, as in the BIDS specs).
- d4b6567f578f0c86ecbb153926c1f178c47cae73 where the documentation is extended

effigies:
I believe the test failure is that `sd_syn_sloppy.json` hasn't been updated to match the number of levels of `sd_syn.json`.

sdcflows/data/sd_syn_sloppy.json
```diff
@@ -1,25 +1,25 @@
...
+    "convergence_threshold": [ 1e-06, 1e-08 ],
+    "convergence_window_size": [ 5, 2 ],
     "dimension": 3,
     "interpolation": "Linear",
-    "metric": [ "Mattes", "CC", "CC" ],
-    "metric_weight": [ 1, 1, 1 ],
-    "number_of_iterations": [ [ 10, 0 ], [ 10, 5, 0 ], [ 2 ] ],
+    "metric": [ ["Mattes", "Mattes"], ["Mattes", "Mattes"] ],
+    "metric_weight": [ [0.5, 0.5], [0.5, 0.5] ],
+    "number_of_iterations": [ [ 20, 10 ], [ 2 ] ],
```
effigies, line 9: Same as `sd_syn`, but with factor of 0.1 and 0.2 applied, matching the previous factors 0.1, 0.5, 0.2.

oesteban:
Okay, I think all the comments have been addressed - I'm going to press the red (green) button now.

PR #243, opened 2021-10-14
DOC: Self-hosted & multiversion documentation overhaul

oesteban:
Following after nipreps/niworkflows#656


oesteban:
Pending: updating `gh-pages` accordingly - will do later

docs/Makefile
```diff
@@ -53,10 +53,13 @@ clean:
 	rm -rf $(BUILDDIR)/*
 	rm -rf reference/*
 
+docs-coverage:
+	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b coverage $(ALLSPHINXOPTS) $(BUILDDIR)/coverage
+
 html:
-	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	PYTHONPATH=$(PYTHONPATH) $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/$(CURBRANCH)
```
effigies: This is going to make it trickier to rerender older versions at need. Or rather, it will create a watershed where we do one thing to render before and another thing to render after. Even if we can't get a single "correct" result now, I would suggest that we leave this rule unchanged and create a new rule like `htmlbranch`.

oesteban: Since deployment is now decoupled (one needs to actively checkout `gh-pages` and add/edit contents) I don't think it has much of an effect.

What probably would make sense is to change the name of the variable - OUTDIR would be much more adequate, and the default value would be `"html"` for consistency with the former `make html` target. WDYT?

PR #247, opened 2021-10-26
FIX: Implement ``B0FieldIdentifier`` / ``B0FieldSource``

### LINKED ISSUE #246: Incorrect handling of new BIDS' ``B0FieldIdentifier``
Trying to estimate only once from two pairs of PEPOLAR fieldmaps (sharing shimming parameters and echo timings), it seems that `find_fieldmaps` does not follow the `B0FieldIdentifier`/`B0FieldSource` very closely:

```
Process Process-2:
  File "/home/oesteban/.miniconda/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/oesteban/.miniconda/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/data/home/oesteban/workspace/dmriprep/dmriprep/cli/workflow.py", line 107, in build_workflow
    retval["workflow"] = init_dmriprep_wf()
  File "/data/home/oesteban/workspace/dmriprep/dmriprep/workflows/base.py", line 80, in init_dmriprep_wf
    single_subject_wf = init_single_subject_wf(subject_id)
  File "/data/home/oesteban/workspace/dmriprep/dmriprep/workflows/base.py", line 335, in init_single_subject_wf
    fmap_estimators = find_estimators(
  File "/data/home/oesteban/workspace/sdcflows/sdcflows/utils/wrangler.py", line 254, in find_estimators
    e = fm.FieldmapEstimation(
  File "<attrs generated init sdcflows.fieldmaps.FieldmapEstimation>", line 7, in __init__
    self.__attrs_post_init__()
  File "/data/home/oesteban/workspace/sdcflows/sdcflows/fieldmaps.py", line 429, in __attrs_post_init__
    _estimators[self.bids_id] = self.paths()
  File "/data/home/oesteban/workspace/sdcflows/sdcflows/utils/bimap.py", line 144, in __setitem__
    raise KeyError(
KeyError: "'topup_001' is already  in mapping"
```

oesteban:
Start with the implementation of the new BIDS metadata entries, to
facilitate more reliable estimation and correction with fieldmaps.

Resolves: #246.


sdcflows/utils/wrangler.py
```diff
@@ -232,57 +258,74 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
...
-        )
-        estimators.append(e)
-
-    # A bunch of heuristics to select EPI fieldmaps
-    sessions = layout.get_sessions()
-    acqs = tuple(layout.get_acquisitions(suffix="epi") + [None])
-    contrasts = tuple(layout.get_ceagents(suffix="epi") + [None])
+    b0_ids = tuple()
+    with suppress(BIDSEntityError):
+        b0_ids = layout.get_B0FieldIdentifiers()
```
mgxd: TIL about this convenience function. We probably want to ensure we are only concerned with the current subject.

```suggestion
        b0_ids = layout.get_B0FieldIdentifiers(**base_entities)
```

sdcflows/utils/wrangler.py
```diff
@@ -232,57 +258,74 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
...
-            for intent in epi_fmap.get_metadata()["IntendedFor"]
-        ]
-
-        epi_sources = []
-        for fmap in targets:
-            with suppress(fm.MetadataError):
-                epi_sources.append(
-                    fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+        # A bunch of heuristics to select EPI fieldmaps
+        sessions = layout.get_sessions()
```
mgxd:
```suggestion
        sessions = layout.get_sessions(subject=subject)
```

sdcflows/utils/wrangler.py
```diff
@@ -232,57 +258,74 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
...
+        acqs = tuple(layout.get_acquisitions(suffix="epi") + [None])
```
mgxd:
```suggestion
        acqs = tuple(layout.get_acquisitions(subject=subject, suffix="epi") + [None])
```

sdcflows/utils/wrangler.py
```diff
@@ -232,57 +258,74 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
...
+        contrasts = tuple(layout.get_ceagents(suffix="epi") + [None])
```
mgxd:
```suggestion
        contrasts = tuple(layout.get_ceagents(subject=subject, suffix="epi") + [None])
```

sdcflows/utils/wrangler.py
```diff
@@ -295,6 +338,8 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
 
     from .epimanip import get_trt
 
+    # Sessions may not be defined at this point if some id was found.
+    sessions = layout.get_sessions()
```
mgxd:
```suggestion
    sessions = layout.get_sessions(subject=subject)
```

mgxd (commented):
I suggested some additional query arguments to ensure we only grab subject relevant files.

Will this require us to make changes to the [current fmriprep / nibabies way of handling the `B0*` metadata](https://github.com/nipreps/fmriprep/blob/08c1e4e1dd2374b5dde777b51e7b9b5cf8d9ec01/fmriprep/workflows/bold/base.py#L246-L247)?

sdcflows/utils/wrangler.py
```diff
@@ -232,57 +258,74 @@ def find_estimators(*, layout, subject, fmapless=True, force_fmapless=False):
...
 
-    # Set up B0 fieldmap strategies:
-    for fmap in layout.get(suffix=["fieldmap", "phasediff", "phase1"], **base_entities):
-        e = fm.FieldmapEstimation(
-            fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
-        )
-        estimators.append(e)
+    b0_ids = tuple()
+    with suppress(BIDSEntityError):
+        b0_ids = layout.get_B0FieldIdentifiers(**base_entities)
```
effigies, line 263: Do we want to specify `extension` and `scope`? I think only `subject` is relevant. Possibly `session`.

oesteban: I don't think `extension` would be required, yes. But, in the case of a weird dataset where pybids finds a json without its NIfTI this would fail for very little gain.

`scope` is necessary, because `derivatives/` will definitely have `B0FieldIdentifier` defined if processed with *SDCFlows*.

`subject` is necessary, of course,

and `session` should not be there - the user may want to use fieldmaps with data from another session and save time in others (e.g., My Connectome).

effigies: Okay, as long as extension doesn't break things, that's fine.

effigies: > `scope` is necessary, because `derivatives/` will definitely have `B0FieldIdentifier` defined if processed with _SDCFlows_.

But suppose I have a qMRI sequence where I quantify B0, which would be a derivative, and want to use that to correct raw BOLD series?

oesteban: If it is preprocessed, and in Hz, then you don't need an estimator for that - this function finds data that potentially can be used to estimate a fieldmap.

If, on the contrary, some preprocessing is needed (e.g., you will want to approximate some B-Spline coefficients to properly integrate with the correction workflows), then *SDCFlows* does not support that type of fieldmap. You would definitely pick up some `sdcflows.interfaces` but essentially, you have to build the workflow yourself. Also, our `Fieldmap` and `FieldmapEstimator` objects do not support other types of fieldmaps.

oesteban: The responsibility of this function is merely implementing the best we can those heuristics we used to have in fMRIPrep, for convenience across the board (dMRIPrep, babies, rodents, etc.).

oesteban: Before I merge, I guess I didn't explicitly give an answer to the actual question.

> But suppose I have a qMRI sequence where I quantify B0, which would be a derivative, and want to use that to correct raw BOLD series?

Then, (i) you want it to have a ``B0FieldIdentifier`` in it, or at the least the ``IntendedFor`` so that [the code to find fieldmap matches locates it](https://github.com/nipreps/fmriprep/blob/08c1e4e1dd2374b5dde777b51e7b9b5cf8d9ec01/fmriprep/workflows/bold/base.py#L246-L272), (ii) if you want to use *SDCFlows*' unwarp workflow, then you need to make sure that an anatomical reference is given and B-Spline coefficients representation available - if not, that's something you'll write on your own, (iii) [invoke the unwarp workflow](https://github.com/nipreps/fmriprep/blob/08c1e4e1dd2374b5dde777b51e7b9b5cf8d9ec01/fmriprep/workflows/bold/base.py#L1031-L1191).

As you will see with the links, both (i) and (iii) take place within the downstream library (i.e., not implemented in *SDCFlows* proper).

PR #251, opened 2021-11-24
FIX: Mishandling of ``topup`` coefficients with higher resolution EPIs

### LINKED ISSUE #250: (2.0.9) bspline.py: error due to incorrect expected shapes of coefficient vs. reference images when knot spacing = 1 voxel?
(Tagging #154 as this appears to be related.)

In **bspline.py**, _fix_topup_fieldcoeff assumes the fieldcoeff map output by FSL topup is padded by 3 voxels:

```
    coeff_shape = np.array(coeffnii.shape[:3])
    ref_shape = np.array(refnii.shape[:3])
    factors = coeffnii.header.get_zooms()[:3]
    if not np.all(coeff_shape == ref_shape // factors + 3):
        raise ValueError(
            f"Shape of coefficients file {coeff_shape} does not meet the "
            f"expectation given the reference's shape {ref_shape}."
        )
```
But the 3-voxel padding does not seem to be applied when the knot spacing is one voxel, resulting in an error due to discrepancy between expected and actual shapes of the coeff and ref volumes (see error below). Assuming the documentation of `--warpres` at 

http://ftp.nmr.mgh.harvard.edu/pub/dist/freesurfer/tutorial_packages/centos6/fsl_507/doc/wiki/topup(2f)TopupUsersGuide.html 

is accurate for FSL 5.0.11, this would occur when the smallest dimension of the input image (`--imain`) is greater than 1/2 the final knot spacing (`--warpres`).

On fMRIPrep 21.0.0rc2 I have had this error with 3.0 and 4.0 mm pepolar fmaps, but not 1.6 and 2.0 mm fmaps. (Final warpres in b02b0.cnf is 4 mm).

For the below error, the reference fmap  was 74 x 74 x 48, 3.0 mm isotropic.
 
```
	 [Node] Finished "fmriprep_wf.single_subject_<SUB>_wf.fmap_preproc_wf.wf_auto_00001.topup".
211119-21:20:15,21 nipype.workflow WARNING:
	 Storing result file without outputs
211119-21:20:15,28 nipype.workflow WARNING:
	 [Node] Error on "fmriprep_wf.single_subject_<SUB>_wf.fmap_preproc_wf.wf_auto_00001.fix_coeff" (/wd/fmriprep_wf/single_subject_<SUB>_wf/fmap_preproc_wf/wf_auto_00001/fix_coeff)
211119-21:20:15,31 nipype.workflow ERROR:
	 Node fix_coeff failed to run on host <HOST>.
211119-21:20:15,39 nipype.workflow ERROR:
	 Saving crash info to /out/sub-<SUB>/log/20211119-201526_b0f73562-529f-49e7-9cb4-9e16ec95cae1/crash-20211119-212015-tmadison-fix_coeff-71d8433f-782f-4962-aa67-4224830ba263.txt
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/plugins/multiproc.py", line 344, in _send_procs_to_workers
    self.procs[jobid].run(updatehash=updatehash)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 428, in run
  File "/opt/conda/lib/python3.8/site-packages/sdcflows/interfaces/bspline.py", line 420, in _run_interface
    self._results["out_coeff"] = [
  File "/opt/conda/lib/python3.8/site-packages/sdcflows/interfaces/bspline.py", line 422, in <listcomp>
    _fix_topup_fieldcoeff(
  File "/opt/conda/lib/python3.8/site-packages/sdcflows/interfaces/bspline.py", line 480, in _fix_topup_fieldcoeff
    raise ValueError(
ValueError: Shape of coefficients file [74 74 48] does not meet the expectation given the reference's shape [74 74 48].
```

I'm not especially well-versed in NiPreps or FSL, so I would greatly appreciate confirmation whether this is an accurate diagnosis of the issue. Thanks!

oesteban:
The way TOPUP calculates the grid of control points changes for voxels larger than 2x the ``--warpres`` parameter (4mm in our default settings).

Resolves: #250.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -476,7 +476,7 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, refpe_reversed=False, out_file=Non
     coeff_shape = np.array(coeffnii.shape[:3])
     factors = np.array(coeffnii.header.get_zooms()[:3])
     ref_shape = np.array(refnii.shape[:3])
-    exp_shape = ref_shape // factors + 3 if np.all(factors >= 1.0) else ref_shape
+    exp_shape = ref_shape // factors + 3 if np.all(factors > 1.0) else ref_shape
     if not np.all(coeff_shape == exp_shape):
         raise ValueError(
             f"Shape of coefficients file {coeff_shape} does not meet the "
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -474,12 +474,13 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, refpe_reversed=False, out_file=Non
     refnii = nb.load(fmap_ref)
 
     coeff_shape = np.array(coeffnii.shape[:3])
+    factors = np.array(coeffnii.header.get_zooms()[:3])
     ref_shape = np.array(refnii.shape[:3])
-    factors = coeffnii.header.get_zooms()[:3]
-    if not np.all(coeff_shape == ref_shape // factors + 3):
+    exp_shape = ref_shape // factors + 3 if np.all(factors >= 1.0) else ref_shape
+    if not np.all(coeff_shape == exp_shape):
         raise ValueError(
             f"Shape of coefficients file {coeff_shape} does not meet the "
-            f"expectation given the reference's shape {ref_shape}."
+            f"expected shape {exp_shape} (toupup factors are {factors})."
         )
     newaff = np.eye(4)
     newaff[:3, :3] = refnii.affine[:3, :3] * factors
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -476,7 +476,7 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, refpe_reversed=False, out_file=Non
     coeff_shape = np.array(coeffnii.shape[:3])
     factors = np.array(coeffnii.header.get_zooms()[:3])
     ref_shape = np.array(refnii.shape[:3])
-    exp_shape = ref_shape // factors + 3 if np.all(factors > 1.0) else ref_shape
+    exp_shape = ref_shape // factors + 3 * (factors > 1)
     if not np.all(coeff_shape == exp_shape):
         raise ValueError(
             f"Shape of coefficients file {coeff_shape} does not meet the "


sdcflows/interfaces/bspline.py
```diff
@@ -474,12 +474,13 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, refpe_reversed=False, out_file=Non
     refnii = nb.load(fmap_ref)
 
     coeff_shape = np.array(coeffnii.shape[:3])
+    factors = np.array(coeffnii.header.get_zooms()[:3])
     ref_shape = np.array(refnii.shape[:3])
-    factors = coeffnii.header.get_zooms()[:3]
-    if not np.all(coeff_shape == ref_shape // factors + 3):
+    exp_shape = ref_shape // factors + 3 if np.all(factors >= 1.0) else ref_shape
```
oesteban:
```suggestion
    exp_shape = ref_shape // factors + 3 if np.all(factors > 1.0) else ref_shape
```

sdcflows/interfaces/bspline.py
```diff
@@ -474,12 +474,13 @@ def _fix_topup_fieldcoeff(in_coeff, fmap_ref, refpe_reversed=False, out_file=Non
     refnii = nb.load(fmap_ref)
 
     coeff_shape = np.array(coeffnii.shape[:3])
+    factors = np.array(coeffnii.header.get_zooms()[:3])
     ref_shape = np.array(refnii.shape[:3])
-    factors = coeffnii.header.get_zooms()[:3]
-    if not np.all(coeff_shape == ref_shape // factors + 3):
+    exp_shape = ref_shape // factors + 3 if np.all(factors > 1.0) else ref_shape
```
effigies: The title says "larger than 2mm", this says `np.all(factors > 1)`. I'm having trouble squaring either choice with the docstring:

```
    TOPUP-generated "fieldcoeff" files are just B-Spline fields, where the shape
    of the field is fixated to be a decimated grid of the original image by an
    integer factor and added 3 pixels on each dimension.
```

Given that this seems to correspond to choice of `--warpres` passed to `topup` (apologies if I'm misreading the conversation), I wonder if we should take that value as a parameter here. Making the calculations explicit would be much clearer, IMO.

oesteban: The title is definitely inaccurate - it's about the integer factors that determine how far apart control knots are w.r.t. the original image's resolution.

As per the documentation:

> --warpres
> 
> Specifies the resolution (in mm) of the warps. Specifying e.g. --warpres=10 means that an isotropic knot-spacing of 10mm is used for all levels of --subsamp. If instead specifying e.g. --warpres=10,5 indicates that a (n isotropic) knot-spacing of 10mm should be used for the first level of --subsamp followed by 5mm for the second level. A warp resolution of e.g. 10mm does not imply that 10mm is the highest accuracy that can be obtained for registration of any given structure. It is rather related to how fast the field can change as on goes from one point to the next in the field. The warps/field are implemented as cubic/quadratic B-splines with a knot-spacing that has to be an integer multiple of the voxel-size of the --imain images. If a value is specified for --warpres that is not and integer multiple the value will be rounded down. So if for example the voxel-size of the --imain image is 3x3x4mm and --warpes is specified as --warpres=10 the actual resolution (knot-spacing) of the warps will be 9x9x8mm.
> 
> When running topup it is advantageous to go to a quite high warp-resolution, typically all the way to the voxel size (i.e. one spline per voxel). To achieve an actual knot-spacing of 1 voxel in all directions one should specify a knot-spacing smaller than two voxels in the direction with the smallest voxel size. If one e.g. has isotropic voxels with size 2.5x2.5x3.5mm, specifying --warpres=4 (4<2*2.5) will give an actual knot-spacing of 2.5x2.5x3.5mm (i.e. 1x1x1voxels).
> 
> Increasing the resolution means that topup will need more execution time and more memory. Most of the time and memory is spent on estimating and representing the Hessian. It may therefore be advantageous to swith to an alternative minimization method (i.e. Scaled Conjugate Gradient, specified by --minmet=1) for the final high resolution steps.

The resulting factors are then encoded as the zooms in the coefficients NIfTI file.

> Given that this seems to correspond to the choice of ``--warpres`` passed to ``topup`` (apologies if I'm misreading the conversation), I wonder if we should take that value as a parameter here. Making the calculations explicit would be much clearer, IMO.

So the decision `topup` makes internally is encoded on the size and spacings of the coefficients NIfTI file we get at the output, and as @Lestropie pointed out in the linked issue, `topup` is indeed running.

Then, as @madisoth reported, when these factors are 1x1x1 (and I'm not sure it is possible to have anisotropic factors when one of them is 1, as per `--warpres` documentation), we have that the coefficients matrix is not added 3 extra pixels to better cover the FoV with fully supported B-Splines.

effigies: I see. I think my brain was not in a place to fully parse that docstring yesterday. This makes much more sense now.

I think we probably can achieve anisotropic factors with some 1 based on the docstring. For example, `--warpres 6` with voxels `(2.5, 2.5, 3.5)` should result in `(2, 2, 1)`. Note that they say that it should be `<2*min(zoom)` to achieve isotropic 1, not `<2*max(zoom)`.

So I think what you want is either:

```suggestion
    exp_shape = ref_shape // factors + 3 if np.any(factors > 1) else ref_shape
```

Or possibly

```suggestion
    exp_shape = ref_shape // factors + 3 * (factors > 1)
```

Depending on whether they pad in case of any scaling or only on scaled directions.

In both cases I've switched to a integer 1 for the sake of making it clearer that it's an integer comparison. You might consider making `factors = np.uint8(coeffnii.header.get_zooms()[:3])` to be even more explicit.

oesteban: As per this bit of the topup documentation:

> To achieve an actual knot-spacing of 1 voxel in all directions one should specify a knot-spacing smaller than two voxels in the direction with the smallest voxel size. If one e.g. has isotropic voxels with size 2.5x2.5x3.5mm, specifying --warpres=4 (4<2*2.5) will give an actual knot-spacing of 2.5x2.5x3.5mm (i.e. 1x1x1voxels).

I'm unsure the factors will not be all ones when the smallest zoom is larger than half warpres.

oesteban: That said, your second suggestion should work either way

effigies (approved):

effigies:
Is this waiting on any confirmation or should we get it in?

oesteban:
I think it can go in.

PR #258, opened 2022-01-13
FIX: Create one fieldmap estimator per EPI-IntendedFor pair

effigies:

PR #259, opened 2022-01-14
ENH: Add max_trs parameter to PEPOLAR workflows, set default to 5

effigies:
From https://github.com/nipreps/sdcflows/pull/258#issuecomment-1013108359.

TOPUP will gladly use as many volumes as we give it, and does not simply find an average volume in each direction. So 50 volumes is quite a lot. In no cases do people acquiring PEPOLAR fieldmaps actually acquire 50 volumes. The most observed in OpenNeuro is 20, and 85% of datasets use 5 or fewer.

By setting the default to 5, we match the likely intent of a majority of observed cases. By adding it as a parameter, we allow downstream tools like fMRIPrep and dMRIPrep to choose (or provide a CLI parameter for users to choose) a more appropriate number.

Changing the default value may violate our bug-fix rules, so that might need to go in a new minor series. But adding the `max_trs` input could work in bug-fix, as leaving it undefined should leave behavior unchanged. That would allow fMRIPrep 21.0 to give users the ability to use BOLD files without requiring useless hours of processing.

### DIFF
--- a/sdcflows/interfaces/utils.py
+++ b/sdcflows/interfaces/utils.py
@@ -51,7 +51,7 @@ class _FlattenInputSpec(BaseInterfaceInputSpec):
         mandatory=True,
         desc="list of metadata",
     )
-    max_trs = traits.Int(50, usedefault=True, desc="only pick first TRs")
+    max_trs = traits.Int(5, usedefault=True, desc="only pick first TRs")
 
 
 class _FlattenOutputSpec(TraitedSpec):
--- a/sdcflows/workflows/fit/pepolar.py
+++ b/sdcflows/workflows/fit/pepolar.py
@@ -27,7 +27,7 @@ from nipype.interfaces import utility as niu
 
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 
-INPUT_FIELDS = ("metadata", "in_data")
+INPUT_FIELDS = ("metadata", "in_data", "max_trs")
 _PEPOLAR_DESC = """\
 A *B<sub>0</sub>*-nonuniformity map (or *fieldmap*) was estimated based on two (or more)
 echo-planar imaging (EPI) references """
@@ -71,6 +71,8 @@ def init_topup_wf(
     metadata : :obj:`list` of :obj:`dict`
         A list of dictionaries containing the metadata corresponding to each file
         in ``in_data``.
+    max_trs : :obj:`int`
+        Maximum number of volumes from each blip to use in estimation.
 
     Outputs
     -------
@@ -148,7 +150,8 @@ def init_topup_wf(
     # fmt: off
     workflow.connect([
         (inputnode, flatten, [("in_data", "in_data"),
-                              ("metadata", "in_meta")]),
+                              ("metadata", "in_meta"),
+                              ("max_trs", "max_trs")]),
         (flatten, readout_time, [("out_data", "in_file"),
                                  ("out_meta", "metadata")]),
         (flatten, regrid, [("out_data", "in_data")]),
@@ -244,6 +247,11 @@ def init_3dQwarp_wf(omp_nthreads=1, debug=False, name="pepolar_estimate_wf"):
     ------
     in_data : :obj:`list` of :obj:`str`
         A list of two EPI files, the first of which will be taken as reference.
+    metadata : :obj:`list` of :obj:`dict`
+        A list of dictionaries containing the metadata corresponding to each file
+        in ``in_data``.
+    max_trs : :obj:`int`
+        Maximum number of volumes from each blip to use in estimation.
 
     Outputs
     -------
@@ -269,9 +277,7 @@ def init_3dQwarp_wf(omp_nthreads=1, debug=False, name="pepolar_estimate_wf"):
 with `3dQwarp` (@afni; AFNI {''.join(['%02d' % v for v in afni.Info().version() or []])}).
 """
 
-    inputnode = pe.Node(
-        niu.IdentityInterface(fields=["in_data", "metadata"]), name="inputnode"
-    )
+    inputnode = pe.Node(niu.IdentityInterface(fields=INPUT_FIELDS), name="inputnode")
 
     outputnode = pe.Node(
         niu.IdentityInterface(fields=["fmap", "fmap_ref"]), name="outputnode"
@@ -344,7 +350,8 @@ with `3dQwarp` (@afni; AFNI {''.join(['%02d' % v for v in afni.Info().version() 
     # fmt: off
     workflow.connect([
         (inputnode, flatten, [("in_data", "in_data"),
-                              ("metadata", "in_meta")]),
+                              ("metadata", "in_meta"),
+                              ("max_trs", "max_trs")]),
         (flatten, sort_pe, [("out_list", "inlist")]),
         (sort_pe, qwarp, [("qwarp_args", "args")]),
         (sort_pe, merge_pes, [("sorted", "in_files")]),
--- a/sdcflows/workflows/fit/pepolar.py
+++ b/sdcflows/workflows/fit/pepolar.py
@@ -104,6 +104,8 @@ def init_topup_wf(
 """
 
     inputnode = pe.Node(niu.IdentityInterface(fields=INPUT_FIELDS), name="inputnode")
+    # Set default so not overridden by `Undefined`. Can be overridden by connection.
+    inputnode.inputs.max_trs = 5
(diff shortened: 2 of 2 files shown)


effigies:
Not really understanding why we're seeing the error we are... @mgxd @oesteban could you have a look?

```
INFO     nipype.workflow:nodes.py:442 [Node] Setting-up "fmap_preproc_wf.wf_auto_00001.flatten" in "/work/fmap_preproc_wf/wf_auto_00001/flatten".
INFO     nipype.workflow:nodes.py:375 [Node] Outdated cache found for "fmap_preproc_wf.wf_auto_00001.flatten".
INFO     nipype.workflow:nodes.py:739 [Node] Running "flatten" ("sdcflows.interfaces.utils.Flatten")
WARNING  nipype.workflow:utils.py:234 Storing result file without outputs
WARNING  nipype.workflow:nodes.py:518 [Node] Error on "fmap_preproc_wf.wf_auto_00001.flatten" (/work/fmap_preproc_wf/wf_auto_00001/flatten)
ERROR    nipype.workflow:tools.py:53 Node flatten failed to run on host 35d6274ac8fc.
ERROR    nipype.workflow:tools.py:70 Saving crash info to /src/sdcflows/crash-20220118-173807-root-flatten-ab21e773-69b1-4e8c-9448-da45737df86d.txt
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 407, in optimize_slicer
    slicer = int(slicer)  # casts float to int as well
TypeError: int() argument must be a string, a bytes-like object or a number, not 'slice'

During handling of the above exception, another exception occurred:

  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/plugins/linear.py", line 45, in run
    node.run(updatehash=updatehash)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 516, in run
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 635, in _run_interface
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/nodes.py", line 741, in _run_command
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 428, in run
  File "/src/sdcflows/sdcflows/interfaces/utils.py", line 76, in _run_interface
    self._results["out_list"] = _flatten(
  File "/src/sdcflows/sdcflows/interfaces/utils.py", line 324, in _flatten
    splitnii = nb.four_to_three(img.slicer[:, :, :, :max_trs])
  File "/opt/conda/lib/python3.8/site-packages/nibabel/spatialimages.py", line 345, in __getitem__
    dataobj = self.img.dataobj[slicer]
  File "/opt/conda/lib/python3.8/site-packages/nibabel/arrayproxy.py", line 397, in __getitem__
    return self._get_scaled(dtype=None, slicer=slicer)
  File "/opt/conda/lib/python3.8/site-packages/nibabel/arrayproxy.py", line 358, in _get_scaled
    scaled = apply_read_scaling(self._get_unscaled(slicer=slicer), scl_slope, scl_inter)
  File "/opt/conda/lib/python3.8/site-packages/nibabel/arrayproxy.py", line 339, in _get_unscaled
    return fileslice(fileobj,
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 787, in fileslice
    segments, sliced_shape, post_slicers = calc_slicedefs(
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 508, in calc_slicedefs
    read_slicers, post_slicers = optimize_read_slicers(
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 571, in optimize_read_slicers
    read_slicer, post_slicer = optimize_slicer(
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 412, in optimize_slicer
    slicer = fill_slicer(slicer, dim_len)
  File "/opt/conda/lib/python3.8/site-packages/nibabel/fileslice.py", line 230, in fill_slicer
    if stop is not None and stop < 0:
TypeError: '<' not supported between instances of '_Undefined' and 'int'
```

Is it just a matter of bumping some cache?

mgxd:
hm, nothing is jumping out to me...I would guess it's a cache problem, just because `IdentityInterface` Nodes behave strangely.

Side note: adding the crashfiles (i.e. `/src/sdcflows/crash-20220118-173807-root-flatten-ab21e773-69b1-4e8c-9448-da45737df86d.txt`) to the artifacts may help debug this in the future.

effigies:
Do you understand the semantics here?

https://github.com/nipreps/sdcflows/blob/b82b872a9b3ac7e5797aef58f0f3f82a12005ba7/.circleci/config.yml#L247-L277

Seems like I can do `[refresh cache]` to fetch a stored cache from OSF, `[wipe <relpath>]` to remove a relative path from `/tmp/work` or bump the cache version, but it's unclear when to do each thing and how persistent their effect is.

mgxd:
to me, it looks like all you need to do is have `[wipe fmap_preproc_wf]` in the commit message. maybe the cached work dir significantly cuts down processing time, but don't know how updated that is (cc @oesteban)

mattcieslak:
One note - in dMRI you can use any of the b=0 images in topup from the epi fmaps or from the dwi scans themselves. The specific volumes you send in to topup can make a huge difference in how well it performs. Because of this, the DHCP pipelines have a method for picking the "most representative" images from each distortion group to include in topup. Their method is to pick the top N images with the highest average spatial correlation with all the other images in the same distortion group. This should work well for bold images too

effigies:
@mattcieslak Thanks! That sounds like a good improvement for a feature release. Is there an existing Python implementation? Or a reference to any implementation would be a good start.

mattcieslak:
Here's my implementation: https://github.com/PennLINC/qsiprep/blob/6d0a3f8fef94ea77c835403363a570f7647c6a82/qsiprep/interfaces/eddy.py#L54

oesteban:
Okay, I believe you are hitting the issue that held me back from implementing these *sophisticated* fieldmaps.

My gut feeling is that we should be using `init_epi_reference_wf` instead of just `Flatten` (e.g., for SyN https://github.com/nipreps/sdcflows/blob/b82b872a9b3ac7e5797aef58f0f3f82a12005ba7/sdcflows/workflows/fit/syn.py#L482-L485).

This is actually why I had such a strong interest in implementing a solid EPI-reference workflow (`niworkflows.workflows.epi.refmap.init_epi_reference_wf`).

However, I can't remember if I just didn't try the `init_epi_reference_wf` integration for laziness or there was any further reason. Basically, we want to generate an EPI reference per run involved (in this case, a single run).

Does this make sense?

oesteban:
To link that with @mattcieslak's comment. The `init_epi_reference_wf` has an input (`t_masks`) which is a bool list indicating whether the particular volume should be averaged within the reference or not.

The global correlation strategy could be implemented so that it generates this `t_masks` list.

\* it's plural because each EPI input has an associated `t_mask`, so you can feed this with one EPI series (which corresponds to the use case at hand) or many (the example of SyN, which now is calculated only once for all runs).

oesteban:
Regarding the actual error (if you insist to go forward with `Flatten`, which feels to me the wrong path but maybe faster), I don't remember the test-case implementation nor the Flatten implementation with sufficient detail to be of help without digging deeper into it.

mattcieslak:
I think TOPUP doesn't want averaged images but individual examples of images from each distortion group. It will motion correct these images internally and then do the warping

oesteban:
> I think TOPUP doesn't want averaged images but individual examples of images from each distortion group. It will motion correct these images internally and then do the warping

TOPUP will accept these inputs. References are not aligned between them, so TOPUP will account for that naturally. To the eyes of TOPUP this is totally seamless :)

If you have co-registered inputs, though, you can set up topup not to run motion correction.

oesteban:
BTW - this EPI reference (voxelwise median across selected time-steps) is what I meant here https://github.com/nipreps/sdcflows/pull/258#issuecomment-1012841965

because you feed references into TOPUP, it runs with the `fmap/` blip and one additional reference. It doesn't really matter whether you select 5 or 50 TRs to average - they go into TOPUP averaged.

effigies:
IMO this has become more painful than it's worth. Proposed nipreps/fmriprep#2688 to just directly hack in `max_trs` to make this tweakable.

Let's move discussion of filtering volumes and/or pre-merging them to a separate issue (or issues).

PR #260, opened 2022-01-21
REL: 2.0.11

effigies:
In preparation for an fMRIPrep 21.0.1 release.

After this, I will also be creating a `maint/2.0.x` branch so we can start including non-bugfix improvements into `master`.


CHANGES.rst
```diff
@@ -1,3 +1,10 @@
+2.0.11 (January 21, 2022)
```
effigies:
```suggestion
2.0.11 (January 22, 2022)
```

PR #269, opened 2022-05-02
MAINT: Loosen installation restrictions

mgxd:
Since sdcflows is nipreps middleware and will be used by various tools, we should not be overtly strict on installations.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -86,15 +86,16 @@ jobs:
         path: |
           /usr/share/miniconda/pkgs
           /home/runner/.cache/pip
-        key: python-${{ matrix.python-version }}-v2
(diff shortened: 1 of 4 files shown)


mgxd:
i'm not sure why pypy is installed and not cpython... is this just a case of the cache needing to be cleared?

effigies:
We had this issue a couple weeks ago with niworkflows or something. I think the conda-forge builds of Python are now treating cpython and pypy as variant builds of Python, and pypy is lexically later than cpython so is seen as more recent by the resolver.

Sometimes.

mgxd:
I guess clearing the cache worked... good to merge?

effigies (approved):

effigies:
We'll see.

PR #270, opened 2022-05-13
CI: Ensure builds are triggered weekly

mgxd:


effigies:
Is this to keep the cache from expiring? If not, maybe we should drop cache so we're not surprised by mismatches between `master` and PRs?

mgxd:
is there an easy way to specify a stale cache on GHA? I agree dropping the cache can be helpful (just came up in #269) but might be excessive doing it every build

effigies:
My guess would be adding something that changes periodically to the cache name, like `${{ date.week }}` (I just made that up).

effigies:
Well, we can worry about better cache names later. Sorry for dropping the ball on this one, which is clearly a good idea on its own.

PR #274, opened 2022-06-24
ENH: Expose EPI-reference to fieldmap-reference transforms

effigies:
I think this will be necessary for https://github.com/nipreps/fmriprep/issues/2804

cc @lestropie @oesteban

### DIFF
--- a/sdcflows/workflows/apply/registration.py
+++ b/sdcflows/workflows/apply/registration.py
@@ -103,7 +103,7 @@ The field coefficients were mapped on to the reference EPI using the transform.
         name="inputnode",
     )
     outputnode = pe.Node(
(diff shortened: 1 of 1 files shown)


oesteban:
Yes, I agree we want to write out one mapping that allows us to move the positions of the coefficients into the EPI grid(s) we want to correct.

To do that, we want the transform that, with `antsApplyTransforms` (or anything not like FreeSurfer which convention is reversed), resamples the EPI image into the fieldmap reference grid. In other words, this transform maps coordinates in the space of the fieldmap reference grid into the target EPI you want to correct.

oesteban:
@effigies - I've allowed myself to change the title of this PR, I believe the new title reflects the true objective.

That said, I'm not sure we really need this - the output transform file should contain an "unwarping" transform in each BOLD's run space.

PR #278, opened 2022-09-02
FIX: Revise the TOPUP workflow and related utilities

oesteban:
Debugging towards fixing nipreps/fmriprep#2830.

I'm under the impression that the fieldmap estimated by topup is only correct in polarity if the input EPIs are RA{S,I} or AR{S,I} oriented.

The strategy here would be to force all axis to be filled from left to right.

EDIT (new when calling the draft finished):

This PR includes several improvements and changes towards making proper interpretation and reuse of TOPUP outputs:

- Adds code comments at all workflows and functions with substantial edits, trying to help future maintenance.
- Reorients all input images to have positive principal cosines (i.e., axes go L->R, P->A, I->S despite their ordering). The intent is to minimize the possible orientations fed into TOPUP  (see 2444c11).
- Rids TOPUP from calculating head motion, it is done outside, with the hope of a full integration with NiTransforms concatenation of transforms (see 2444c11).
- Revises the ``fix_coeff`` node to consider the fact that the internal coordinate system of FSL is LAS (see e80e68d). I'm rather positive about the LR flip, since I've seen it on both @mgxd's and @Gilles86's data (e.g., TOPUP's *field* and *warpfieldXX* outputs showcase the flip). I'm less positive about the sign change of the field itself (only testable on @Gilles86's data because @mgxd's has *j* direction for PE).
- Reorders the input EPI runs so that TOPUP always gets the same ordering, preferentially starting with a positive direction encoding (i.e., *i*, *j*, or *k*) - see 72640e7

This does not fully-cover nipreps/fmriprep#2830 (~~nor #218~~ EDIT -  it does cover this issue, finally), but I have checked that at least the reference image is generated by *SDCFlows*, and the corresponding by TOPUP are in spatial agreement (i.e., correction is properly applied to the EPIs fed in the estimation) on both datasets (Mathias' and Gilles'). A subsequent PR will address the unwarping of `IntendedFor` EPI targets.


oesteban:
Okay, for the most part, this seems to work on both @Gilles86 and @mgxd datasets - not sure about generalization to new datasets, but this is a start.

There's a remaining issue with @mgxd, being that the registration between the fieldmap reference and the EPI runs is not great. But I think that should be addressed on a separate issue.

Reviews at this point are welcome and important.

cc/ @effigies

sdcflows/interfaces/epi.py
```diff
@@ -64,3 +66,76 @@ def _run_interface(self, runtime):
...
+        desc="list of PE directions, in FSL's conventions",
+    )
+    readout_times = OutputMultiObject(
+        traits.Float,
+        desc="list of total readout times"
+    )
+
+
+class SortPEBlips(SimpleInterface):
+    """Sort PE blips so they are consistently fed into TOPUP."""
```
mgxd, line 109: Could you write up a test for this?

sdcflows/workflows/fit/pepolar.py
```diff
@@ -171,40 +194,38 @@ def init_topup_wf(
...
 
-    from nipype.interfaces.afni.preprocess import Volreg
     from niworkflows.interfaces.nibabel import SplitSeries
     from ...interfaces.bspline import ApplyCoeffsField
 
-    realign = pe.Node(
-        Volreg(args=f"-base {grid_reference}", outputtype="NIFTI_GZ"),
-        name="realign_blips",
-    )
+    # Separate the runs again, as our ApplyCoeffsField corrects them separatelly
```
mgxd: super nitpick but typo
```suggestion
    # Separate the runs again, as our ApplyCoeffsField corrects them separately
```

setup.cfg
```diff
@@ -32,13 +32,13 @@ install_requires =
     nibabel >=3.0.1
     nipype >=1.5.1,<2.0
     traits <6.4
-    niworkflows ~= 1.6.0
+    niworkflows @ git+https://github.com/nipreps/niworkflows.git@master
```
mgxd: I just cut a new version, this can be set to `~= 1.6.3`

setup.cfg
```diff
@@ -32,13 +32,13 @@ install_requires =
...
     nitransforms >= 21.0.0
     numpy >= 1.21.0
-    pybids >= 0.12.1
+    pybids >= 0.15.1
```
mgxd, line 38: Is this needed or can we hold off until a new minor release

sdcflows/interfaces/utils.py
```diff
@@ -297,6 +297,33 @@ def _run_interface(self, runtime):
         return runtime
 
 
+class _PositiveDirectionCosinesInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True, desc="input image")
+
+
+class _PositiveDirectionCosinesOutputSpec(TraitedSpec):
+    out_file = File(exists=True)
```
mgxd:
```suggestion
    out_file = File()
```

sdcflows/interfaces/epi.py
```diff
@@ -64,3 +66,76 @@ def _run_interface(self, runtime):
...
+    readout_times = InputMultiObject(
+        traits.Float,
+        mandatory=True,
+        desc="list of total readout times"
+    )
+
+
+class _SortPEBlipsOutputSpec(TraitedSpec):
+    out_data = OutputMultiObject(
+        File(exists=True),
```
mgxd: https://github.com/nipy/nipype/issues/3493
```suggestion
        File(),
```

mgxd (commented):
Did a quick pass, still having trouble conceptualizing the whole process. Do you happen to have some screenshots on how these results look on the troublesome datasets before / after this patch?

sdcflows/interfaces/epi.py
oesteban: Yup

setup.cfg
oesteban: Were we not going to use bleeding edge pinning for dev branches, and limited pinning when about to cut a new release?

setup.cfg
oesteban: Required by niworkflows, actually.

sdcflows/interfaces/utils.py
oesteban: ~~Sure, although this has no effect, as we were discussing in niworkflows.~~

setup.cfg
mgxd: I think we will be instantly cutting a release upon merge if this fixes the problems we were seeing.

oesteban:
```suggestion
    niworkflows ~= 1.6.3
```

oesteban:
Okay, now with tests revised for the new set of reorientation operations

sdcflows/interfaces/epi.py
```diff
@@ -64,3 +66,76 @@ def _run_interface(self, runtime):
...
+        desc="list of total readout times"
+    )
+
+
+class _SortPEBlipsOutputSpec(TraitedSpec):
+    out_data = OutputMultiObject(
+        File(),
+        desc="list of input data",
+    )
+    pe_dirs = InputMultiObject(
```
mgxd:
```suggestion
    pe_dirs = OutputMultiObject(
```

oesteban:
> still having trouble conceptualizing the whole process. Do you happen to have some screenshots on how these results look on the troublesome datasets before / after this patch?

I had @Gilles86's reports - maybe he could confirm whether things are looking better now? In his case (nipreps/fmriprep#2830), the displacements were flipped (i.e., in the opposed direction).

On your data, @mgxd, the diagnosis was similar. However, the results are still a bit off because (I suspect but haven't confirmed) of the registration between the distorted EPI and fieldmap reference.

Still hoping to put together more visual documentation to make it easier to understand both the problem and the solution.

oesteban:
I'll merge after @eilidhmacnicol's blessings.

sdcflows/interfaces/bspline.py
```diff
@@ -448,36 +457,63 @@ def bspline_grid(img, control_zooms_mm=DEFAULT_ZOOMS_MM):
...
     refnii = nb.load(fmap_ref)
 
+    # Load coefficients
+    coeffnii = nb.load(in_coeff)
+    # Coefficients - flip LR and overwrite coeffnii variable
+    # Internal data orientation of FSL is LAS, so coefficients will be LR flipped,
+    # and because the affine does not encode orientation (factors instead), this flip
+    # always is implicit.
+    # If the PE direction is x/i, the flip in the axis direction causes that the
+    # fieldmap estimation must also be inverted in direction (multiply by -1.0)
```
eilidhmacnicol, line 480: Just checking that I understand this correctly: this would have no impact on an image if the PE direction is IS/SI (as is the case in most of our images) or AP/PA, because the fieldmap estimation does not need to be flipped even if the data orientation is?

sdcflows/transform.py
```diff
@@ -94,9 +103,9 @@ def fit(self, spatialimage):
         )
 
         # Cache
-        self.shifts = nb.Nifti1Image(vsm, spatialimage.affine, None)
-        self.shifts.header.set_intent("estimate", name="Voxel shift")
-        self.shifts.header.set_xyzt_units("mm")
+        self.mapped = nb.Nifti1Image(vsm, spatialimage.affine, None)
+        self.mapped.header.set_intent("estimate", name="Voxel shift")
+        self.mapped.header.set_xyzt_units("mm")
```
eilidhmacnicol: Not an issue now, but may need to come back to this if we plan to change units to microns for small voxel sizes. For example, would it be better to set the units based on `spatialimage.header`?

eilidhmacnicol (commented):
LGTM. Left a few comments, but nothing critical to address before merging.

What was the rationale behind the `b02b0.cnf` changes? I understand the `estmov` change to stop the estimation of motion (since it is done elsewhere) but what was the reasoning for the reduction of levels to calculate the field?

sdcflows/interfaces/bspline.py
oesteban: Correct

sdcflows/transform.py
oesteban: Yes, it would be better to pick it from `spatialimage.header`. I believe the problem was that `spatialimage` may not be available sometimes, but it could be just laziness.

oesteban: Yes, actually laziness:
```suggestion
        self.mapped.header.set_xyzt_units(*spatialimage.header.get_xyzt_units())
```

oesteban:
> What was the rationale behind the `b02b0.cnf` changes? I understand the `estmov` change to stop the estimation of motion (since it is done elsewhere) but what was the reasoning for the reduction of levels to calculate the field?

I could have kept the number of levels, but it didn't make much sense as the blips come already aligned. Keeping them was a waste of resources and time for no improvement of the nonlinear component of the estimation (which is done in the latter levels of registration).

sdcflows/transform.py
```diff
@@ -94,9 +103,9 @@ def fit(self, spatialimage):
         )
 
         # Cache
-        self.shifts = nb.Nifti1Image(vsm, spatialimage.affine, None)
-        self.shifts.header.set_intent("estimate", name="Voxel shift")
-        self.shifts.header.set_xyzt_units("mm")
+        self.mapped = nb.Nifti1Image(vsm, spatialimage.affine, None)
+        self.mapped.header.set_intent("estimate", name="Voxel shift")
+        self.mapped.header.set_xyzt_units(spatialimage.header.get_xyzt_units())
```
oesteban:
```suggestion
        self.mapped.header.set_xyzt_units(*spatialimage.header.get_xyzt_units())
```

Gilles86:
Would it be possible to send me the `_boldref.nii.gz`-files?

PR #284, opened 2022-09-26
ENH: Make wrangler more verbose

### LINKED ISSUE #283: Improve debug verbosity of `find_estimators`
There is a lot going on, which makes it hard to debug issues such as https://github.com/nipreps/sdcflows/issues/286

We could add a debug parameter that would allow displaying more information.

mgxd:
Closes #283 

Adds `verbose` parameter, which enables logging through the `find_estimators()` process. 

I think this could also benefit from a way to visualize the layout, to further help debug why files are not being found.

### DIFF
--- a/sdcflows/cli/find_estimators.py
+++ b/sdcflows/cli/find_estimators.py
@@ -32,6 +32,12 @@ def _parser():
         help="Path to a PyBIDS database folder, for faster indexing (especially "
              "useful for large datasets). Will be created if not present."
     )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="Print information while finding estimators (Useful for debugging)",
+    )
     return parser
 
 
@@ -76,7 +82,7 @@ def main(argv=None):
     estimators_record = {}
     for subject in subjects:
         estimators_record[subject] = find_estimators(
-            layout=layout, subject=subject, fmapless=pargs.fmapless
+            layout=layout, subject=subject, fmapless=pargs.fmapless, verbose=pargs.verbose,
         )
 
     # pretty print results
--- a/sdcflows/cli/find_estimators.py
+++ b/sdcflows/cli/find_estimators.py
@@ -73,16 +73,21 @@ def main(argv=None):
     """
     from niworkflows.utils.bids import collect_participants
     from sdcflows.utils.wrangler import find_estimators
+    from sdcflows.utils.misc import create_logger
 
     pargs = _parser().parse_args(argv)
 
     bids_dir = pargs.bids_dir.resolve(strict=True)
     layout = gen_layout(bids_dir, pargs.bids_database_dir)
     subjects = collect_participants(layout, pargs.subjects)
+    logger = create_logger('sdcflow.wrangler', level=10 if pargs.verbose else 40)
     estimators_record = {}
     for subject in subjects:
         estimators_record[subject] = find_estimators(
-            layout=layout, subject=subject, fmapless=pargs.fmapless, verbose=pargs.verbose,
+            layout=layout,
+            subject=subject,
+            fmapless=pargs.fmapless,
+            logger=logger,
         )
 
     # pretty print results
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -29,8 +29,8 @@ from bids.layout import BIDSLayout
 from bids.utils import listify
 
 
-
-def find_estimators(*,
+def find_estimators(
+    *,
     layout: BIDSLayout,
     subject: str,
     sessions: Optional[List[str]] = None,
@@ -284,7 +284,10 @@ def find_estimators(*,
         has_B0FI = True
 
     if has_B0FI:
-        logger.debug("Dataset includes B0FieldIdentifier metadata. All `IntendedFor` metadata will be ignored.")
+        logger.debug(
+            "Dataset includes `B0FieldIdentifier` metadata."
+            "Any data missing this metadata will be ignored."
+        )
         for b0_id in b0_ids:
             # Found B0FieldIdentifier metadata entries
             b0_entities = base_entities.copy()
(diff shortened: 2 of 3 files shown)


sdcflows/utils/wrangler.py
```diff
@@ -244,46 +256,68 @@ def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmap
...
+            e = fm.FieldmapEstimation([
+                fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+                for fmap in layout.get(**b0_entities)
+            ])
+            logger.debug(
+                "Found %s estimation from %d sources:\n- %s",
+                e.method.name,
+                len(e.sources),
+                "\n- ".join([str(x.path.relative_to(layout.root)) for x in e.sources]),
+            )
```
effigies: With as many times as you've written exactly this, I think it could be useful to add a function or method that would emit a description:

```Python
def describe(est, root):
    return "\n- ".join(
        [f"Found {est.method.name} estimation from {len(est.sources)} sources:"] +
        [str(src.path.relative_to(root)) for src in est.sources]
    )
```

Then all these debug statements become:

```suggestion
            logger.debug(describe(e, layout.root))
```
or

```suggestion
            logger.debug(e.describe(layout.root))
```

effigies (approved):

sdcflows/utils/wrangler.py
```diff
@@ -243,47 +258,58 @@ def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmap
                        bids_id='auto_00000')]
 
     """
-    from .. import fieldmaps as fm
+    from .misc import create_logger
     from bids.layout import Query
     from bids.exceptions import BIDSEntityError
 
+    # Set up logger (logs only visable if verbose)
+    logger = create_logger('sdcflows.wrangler', level=10 if verbose else 40)
```
oesteban:
```suggestion
    logger = logger or create_logger('sdcflows.wrangler', 40)
```

sdcflows/utils/wrangler.py
```diff
@@ -21,13 +21,26 @@
...
-def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmapless=False):
+
+def find_estimators(
+    *,
+    layout: BIDSLayout,
+    subject: str,
+    sessions: Optional[List[str]] = None,
+    fmapless: Union[bool, set] = True,
+    force_fmapless: bool = False,
+    verbose: bool = False,
```
oesteban:
```suggestion
    logger: logging.Logger = None,
```

sdcflows/utils/wrangler.py
```diff
@@ -243,47 +258,58 @@ def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmap
...
         fmapless = {"bold", "dwi"}
 
     estimators = []
 
-    b0_ids = tuple()
+    # Step 1. Use B0FieldIdentifier metadata
+    has_B0FI = False
     with suppress(BIDSEntityError):
         b0_ids = layout.get_B0FieldIdentifiers(**base_entities)
+        has_B0FI = True
```
oesteban: why this is necessary? Isn't it robuster to just check `bool(b0_ids)` ?

sdcflows/utils/wrangler.py
```diff
@@ -243,47 +258,58 @@ def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmap
...
+
+    if has_B0FI:
```
oesteban:
```suggestion
    if b0_ids:
```

sdcflows/utils/wrangler.py
```diff
@@ -243,47 +258,58 @@ def find_estimators(*, layout, subject, sessions=None, fmapless=True, force_fmap
...
+            # Found B0FieldIdentifier metadata entries
+            b0_entities = base_entities.copy()
+            b0_entities["B0FieldIdentifier"] = b0_id
+
+            e = fm.FieldmapEstimation([
+                fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+                for fmap in layout.get(**b0_entities)
+            ])
+            _log_debug_estimation(logger, e, layout.root)
+            estimators.append(e)
```
oesteban, line 302:
```suggestion
    for b0_id in b0_ids:
        # Found B0FieldIdentifier metadata entries
        entities = base_entities.copy()
        entities["B0FieldIdentifier"] = b0_id

        _e = fm.FieldmapEstimation([
            fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
            for fmap in layout.get(**entities)
        ])
        estimators.append(_e)
```

oesteban: ditto (no problem if you want to change the `_e` variable name though.

sdcflows/cli/find_estimators.py
```diff
@@ -76,7 +82,7 @@ def main(argv=None):
     estimators_record = {}
     for subject in subjects:
         estimators_record[subject] = find_estimators(
-            layout=layout, subject=subject, fmapless=pargs.fmapless
+            layout=layout, subject=subject, fmapless=pargs.fmapless, verbose=pargs.verbose,
```
oesteban:
```suggestion
            layout=layout,
            subject=subject,
            fmapless=pargs.fmapless,
            verbose=create_logger("sdcflows.wrangler", level=10 if pargs.verbose else 40),
```

oesteban (commented):
I would allow *fMRIPrep* and such to pass in their own logger objects.

sdcflows/utils/wrangler.py
mgxd: why deindent this if it is only relevant if `b0_ids` is not empty

oesteban: I'm not super-strong on this one. That said, IMHO it is better to avoid Python indentations when possible. Also, from a PR point of view, the less lines that *need* a change, the better.

mgxd: I'm gonna go ahead with it, on the basis that I think it makes the logic simpler to follow

PR #285, opened 2022-09-29
DOC: Add a notebook about susceptibility distortions

oesteban:
@effigies, @mgxd, @eilidhmacnicol, @celprov - this is a base standpoint for what I want to cover in today's meeting. Please have a look ahead of time. Comments are welcome.

I want to make a second part, and some improvements to this one. Not sure how much of that I will be able to pull out before the meeting today.


celprov:
@oesteban Nice notebook I think it is very clear. Just one comment :  when you plot the fieldmap, could you include the scale of the colormap as a legend to get a feel of what is the magnitude of the B0 variations ?

oesteban:
> scale of the colormap as a legend to get a feel of what is the magnitude of the B0 variations ?

The scale is in arbitrary units, and I would like to avoid getting into that level of detail here. My plan is that the problem of magnitude/units of the field is addressed in a second notebook about estimation of the fieldmap where that is indeed very relevant.

effigies (approved):
Very clear, IMO.

effigies:
Checks passing after syncing with master. @oesteban is this ready to merge?

oesteban:
I think it is - I was intending to append some cells about pile-up/drop-out. However, that can be done at a later time, and this is already valuable.

PR #287, opened 2022-10-06
FIX: Revise debug/sloppy operations of the ``coeff2epi`` workflow

oesteban:
The ``sloppy`` option was misused for debugging (also because the name was misused).

This PR makes a better use of debug/sloppy boolean arguments.


effigies (approved):

effigies:
If backported, #288 should also be included.

PR #292, opened 2022-10-20
ENH: Allow bids filtering during `get()` calls.

mgxd:
This is to bring SDCFlow's file filtering more in-line with approaches taken in other nipreps. To avoid potential collisions during `get()` parameter assigment, base / additional values are first merged into a dictionary, and then passed in as keyword arguments


mgxd:
blocked by #293

effigies (approved):

effigies:
It looks like we've had a bunch of significant changes since 2.1.1, so `master` will probably be a 2.2.0 release. Do you want to target 2.1.2? I've just pushed a `maint/2.1.x` branch.

effigies:
Can backport if needed.

PR #293, opened 2022-10-20
FIX: Ensure `replace()` calls only alter the file basename

mgxd:
This came up while generating a `phasediff` directory to test #292


oesteban (approved):
LGTM

Issue #266, opened 2022-04-09
IntendedFor is ignored by wrangler.find_estimators


bpinsard:
I am trying the new fMRIPrep release 21.0.1, and I have never ending topup on pilot multi-echo data.
Trying to investigate why, I found out that SDCFlow ignored the `IntendedFor`  and topup is run on 20+ incongruent volumes (echoes with different distortions level from all possible epi fieldmap in the session).

Looking at the code, if I am not mistaken, it seems that the `IntendedFor` will never be used as the following code will take all EPI and return a single estimator per session.
https://github.com/nipreps/sdcflows/blob/56a02bfa17a3640e91b4073ba4e5237bce19dcee/sdcflows/utils/wrangler.py#L291-L304 
and then IntendedFor are skipped because the epi fmap is already used above. 
https://github.com/nipreps/sdcflows/blob/56a02bfa17a3640e91b4073ba4e5237bce19dcee/sdcflows/utils/wrangler.py#L313-L314

I would say that this issue is not multi-echo specific as it did not only aggregate echoes from the same series. The BIDS might be faulty as IntendedFor in multi-echo fMRI and phase-reversed EPI fmap is not detailled in the doc (I set it such that the each echo index _epi.json  be IntendedFor the same echo index _bold.nii.gz).
I could make changes to use the new B0FieldIdentifiers BIDS tags, but it would be good if that worked on legacy BIDS datasets using IntendedFor.
Let me know if that's clear or if you need more info on the dataset.

bpinsard:
Looking at the wrangler code, I can see that some options that were working in the previous version might no longer possible.
For instance, running fmriprep on a dataset including fieldmaps with `--use-syn-sdc` to use Syn as fallback when fieldmaps are missing for some of the runs/sessions. 

In addition, from the code logic it seems that if a dataset includes fieldmaps for DWI but not for BOLD, it might require using `--force-syn` to enable Syn on BOLD. As the wrangler is generic for both DWI/BOLD and fmriprep checks if any estimator has been returned, it could interact with the workflow setup. 

https://github.com/nipreps/fmriprep/blob/6ba171f78cc4ece177b68e4953c83bc60c2c9071/fmriprep/workflows/base.py#L335-L341

bpinsard:
Any potential fix for that issue? 
I am also puzzle by the pybids query to find the fmaps tagged with B0FieldIdentifier here
[...]

PR #298, opened 2022-11-10
ENH: Find B0FieldIdentifiers when one image contributes to multiple

bpinsard:
Attempt at fixing/hacking #266


sdcflows/fieldmaps.py
```diff
@@ -404,13 +404,15 @@ def __attrs_post_init__(self):
...
+                for f in self.sources
+                if f.metadata.get("B0FieldIdentifier")
+            ]
+
+            # intersection of B0FieldIdentifiers (in case of lists)
+            if len(bids_ids) > 1:
+                bids_ids = bids_ids[0].intersection(*bids_ids[1:])
             if len(bids_ids) > 1:
                 raise ValueError(
                     f"Multiple ``B0FieldIdentifier`` set: <{', '.join(bids_ids)}>"
```
effigies: We're changing the logic significantly, here and the old check doesn't make sense. What about?

```suggestion
            b0_ids = [
                listify(f.metadata.get("B0FieldIdentifier"))
                for f in self.sources
                if f.metadata.get("B0FieldIdentifier")
            ]
            bids_ids = set().intersection(*b0_ids)
            
            if not bids_ids:
                raise ValueError(
                    f"No common ``B0FieldIdentifier`` found: <{', '.join(b0_ids)}>"
                )
            elif len(bids_ids) > 1:
                raise ValueError(
                    f"Multiple common ``B0FieldIdentifier``s found: <{', '.join(bids_ids)}>"
```

sdcflows/fieldmaps.py
```diff
@@ -404,16 +404,20 @@ def __attrs_post_init__(self):
...
-            )
-            if len(bids_ids) > 1:
+            b0_ids = [
+                listify(f.metadata.get("B0FieldIdentifier"))
+                for f in self.sources
+                if f.metadata.get("B0FieldIdentifier")
+            ]
+            bids_ids = set().intersection(*b0_ids)
+            
+            if not bids_ids:
```
effigies: Right, if we don't find any `B0FieldIdentifier`s, that's a different situation from not finding any common ones.

```suggestion
            if not bids_ids and any(ids for ids in b0_ids):
```

sdcflows/fieldmaps.py
```diff
@@ -404,16 +404,20 @@ def __attrs_post_init__(self):
...
-                    if f.metadata.get("B0FieldIdentifier")
-                ]
-            )
-            if len(bids_ids) > 1:
+            b0_ids = [
+                listify(f.metadata.get("B0FieldIdentifier"))
+                for f in self.sources
+                if f.metadata.get("B0FieldIdentifier")
+            ]
+            bids_ids = set().intersection(*b0_ids)
```
bpinsard: Merged too fast, the intersection of an empty set with anything will always be an empty set.

effigies: Oh, right. Guess we could do:

```suggestion
            all_ids = set().union(*b0_ids)
            bids_ids = all_ids.intersection(*b0_ids)
```

I just find the `set(a[0]).intersection(*a[1:])` kind of offensive, visually.

bpinsard: I understand, is `a[0].intersection(*a)`  less offensive ?

bpinsard: in fact `set.intersection(*a)` works if called on the class and not an empty instance.

sdcflows/fieldmaps.py
```diff
@@ -404,16 +404,20 @@ def __attrs_post_init__(self):
...
-                    if f.metadata.get("B0FieldIdentifier")
-                ]
-            )
-            if len(bids_ids) > 1:
+            b0_ids = [
+                listify(f.metadata.get("B0FieldIdentifier"))
+                for f in self.sources
+                if f.metadata.get("B0FieldIdentifier")
+            ]
+            bids_ids = set.intersection(*b0_ids)
```
effigies: I wanted `b0_ids` to be lists for the sake of reporting, so I changed them from sets.

```suggestion
            bids_ids = set.intersection(*map(set, b0_ids))
```

sdcflows/fieldmaps.py
```diff
@@ -404,27 +404,31 @@ def __attrs_post_init__(self):
...
-            elif bids_ids:
+            b0_ids = [
+                listify(f.metadata.get("B0FieldIdentifier"))
+                for f in self.sources
+                if f.metadata.get("B0FieldIdentifier")
+            ]
+
+            if b0_ids:
+                # Find common IDs
+                bids_ids = set(b0_ids[0]).intersection(*b0_ids[1:])
```
effigies, line 415: There isn't really a better option here. Ah well.

sdcflows/utils/wrangler.py
```diff
@@ -294,9 +297,15 @@ def find_estimators(
             b0_entities = base_entities.copy()
             b0_entities["B0FieldIdentifier"] = b0_id
 
+            bare_ids = layout.get(**base_entities, B0FieldIdentifier=b0_id)
+            listed_ids = layout.get(
+                **base_entities,
+                B0FieldIdentifier=repr(b0_id),
```
effigies: If I understand this correctly, the idea is that we'll have a tuple of `B0FieldIdentifier`s and when we do a regex_search, it will be `str()`-ified.

```
In [2]: b0_id = "abc"

In [3]: b0_ids = ("abc", "def")

In [4]: f'"{b0_id}"' in str(b0_ids)
Out[4]: False

In [5]: repr(b0_id) in str(b0_ids)
Out[5]: True
```

So I think `repr()` is the right transform here. But I don't have a test dataset set up. Do you have one you could try this on?

effigies:
@bpinsard Spent a bit of time on this logic today, and fixed the tests. Please check it on your use case.

effigies:
Just a bump. If this is good to go, it would be nice to get in for a release.

bpinsard:
Great. will test it asap.

bpinsard:
`B0FieldIdentifier=f'"{b0_id}"'`  works but `repr(b0_id)` gives single quote that are not matched, so if we want to use an existing function `json.dumps` is maybe a better choice?

effigies:
Oh, are we doing `json.dumps` somewhere in pybids to store `B0FieldIdentifier`? I didn't see it. What does `layout.get_B0FieldIdentifiers()` look like on your data?

Will push `f'"{b0_id}"'` for now, though.

sdcflows/utils/wrangler.py
effigies:
```suggestion
                B0FieldIdentifier=f'"{b0_id}"',  # Double quotes to match JSON, not Python repr
```

bpinsard:
`layout.get_B0FieldIdentifiers()` will return string and tuples 

```
In [10]: layout.get_B0FieldIdentifiers()
Out[10]: 
['sub_emotion_ses_002_task_emotion_run_01_part_mag_bold',
 'sub_emotion_ses_001_task_emotion_run_02',
 ('sub_emotion_ses_002_task_emotion_run_01_part_mag_bold',
  'sub_emotion_ses_002_task_emotion_run_02_part_mag_bold',
  'sub_emotion_ses_002_task_emotion_run_03_part_mag_bold'),
 'sub_emotion_ses_001_task_emotion_run_03_part_mag_bold',
 'sub_emotion_ses_002_task_emotion_run_04',
 'sub_emotion_ses_002_task_emotion_run_02_part_mag_bold',
 'sub_emotion_ses_002_task_emotion_run_03_part_mag_bold',
 'sub_emotion_ses_001_task_emotion_run_01_part_mag_bold',
 'sub_emotion_ses_001_task_emotion_run_04']
```
But the regexp matching is likely run on the json serialized string in pybids if I remember well (digged into code and sqlite database a while ago).

effigies:
Okay. This should be working. @bpinsard Would you mind giving it one more try while we're waiting on the CI?

bpinsard:
That works! 🎉 

```
In [7]: sdcflows.utils.wrangler.find_estimators(layout=layout, subject='emotion')
Out[7]: 
[FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_001_task_emotion_run_02'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_001_task_emotion_run_01_part_mag_bold'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_001_task_emotion_run_04'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_001_task_emotion_run_03_part_mag_bold'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_002_task_emotion_run_04'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_002_task_emotion_run_03_part_mag_bold'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_002_task_emotion_run_01_part_mag_bold'),
 FieldmapEstimation(sources=<2 files>, method=<EstimatorType.PEPOLAR: 2>, bids_id='sub_emotion_ses_002_task_emotion_run_02_part_mag_bold')]
```

effigies (approved):

PR #301, opened 2022-11-16
ENH: Integrate downsampling in ``BSplineApprox`` when the input is high-res

oesteban:
Adds a new input ``zooms_min`` that will trigger the downsampling of the input field map when it comes with a very high resolution (typically, this is the case for SDC SyN).

Although the input is resampled, the output field is interpolated at the original resolution for debugging purposes.

Downsampling dramatically speeds up approximation. Alternatively, a solution (for the SDC SyN) would be to calculate a heavily dilated mask of the brain and use it to limit the number of voxels that are fit in regression.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -23,7 +23,6 @@
 """Filtering of :math:`B_0` field mappings with B-Splines."""
 from itertools import product
 from pathlib import Path
-from contextlib import suppress
 import numpy as np
 import nibabel as nb
 from nibabel.affines import apply_affine
@@ -152,14 +151,12 @@ class BSplineApprox(SimpleInterface):
         if need_resize:
             from sdcflows.utils.tools import resample_to_zooms
 
-            zooms_min = self.inputs.zooms_min
-
-            with suppress(TypeError):
-                zooms_min = [float(zooms_min)] * 3
+            zooms_min = np.maximum(zooms, self.inputs.zooms_min)
 
             LOGGER.info(
                 "Resampling image with resolution exceeding 'zooms_min' "
-                f"({'x'.join(str(s) for s in zooms)})."
+                f"({'x'.join(str(s) for s in zooms)} → "
+                f"{'x'.join(str(s) for s in zooms_min)})."
             )
             fmapnii = resample_to_zooms(fmapnii, zooms_min)
 
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -41,7 +41,7 @@ from nipype.interfaces.base import (
     OutputMultiObject,
 )
 
-from sdcflows.transform import grid_bspline_weights as gbsw
+from sdcflows.transform import grid_bspline_weights
 
 
 LOW_MEM_BLOCK_SIZE = 1000
@@ -513,9 +513,9 @@ def _collocation_matrix(image, knot_spacing):
         bs_levels.append(level)
 
         weights = (
-            gbsw(image, level)
+            grid_bspline_weights(image, level)
             if weights is None
-            else sparse_vstack((weights, gbsw(image, level)))
+            else sparse_vstack((weights, grid_bspline_weights(image, level)))
         )
 
     return weights.T.tocsr(), bs_levels
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -23,6 +23,7 @@
 """Filtering of :math:`B_0` field mappings with B-Splines."""
 from itertools import product
 from pathlib import Path
+from contextlib import suppress
 import numpy as np
 import nibabel as nb
 from nibabel.affines import apply_affine
@@ -75,8 +76,10 @@ class _BSplineApproxInputSpec(BaseInterfaceInputSpec):
         usedefault=True,
         desc="generate a field, extrapolated outside the brain mask",
     )
-    zooms_min = traits.Float(
-        1.4,
+    zooms_min = traits.Union(
+        traits.Float,
+        traits.Tuple(traits.Float, traits.Float, traits.Float),
+        default_value=1.95,
         usedefault=True,
         desc="limit minimum image zooms, set 0.0 to use the original image",
     )
@@ -148,14 +151,19 @@ class BSplineApprox(SimpleInterface):
         if need_resize:
             from niworkflows.utils.images import resample_by_spacing
 
+            zooms_min = self.inputs.zooms_min
+
+            with suppress(TypeError):
+                zooms_min = [float(zooms_min)] * 3
+
             LOGGER.info(
                 "Resampling image with resolution exceeding 'zooms_min' "
                 f"({'x'.join(str(s) for s in zooms)})."
             )
-            fmapnii = resample_by_spacing(fmapnii, [self.inputs.zooms_min] * 3)
+            fmapnii = resample_by_spacing(fmapnii, zooms_min)
 
             if masknii is not None:
-                masknii = resample_by_spacing(masknii, [self.inputs.zooms_min] * 3)
+                masknii = resample_by_spacing(masknii, zooms_min)
 
         data = fmapnii.get_fdata(dtype="float32")
 
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -170,7 +170,7 @@ class BSplineApprox(SimpleInterface):
 
         # Generate a numpy array with the mask
         mask = (
-            np.ones_like(fmapnii.dataobj, dtype=bool) if masknii is None
+            np.ones(fmapnii.shape, dtype=bool) if masknii is None
             else np.asanyarray(masknii.dataobj) > 1e-4
         )
 
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -47,7 +47,6 @@ LOW_MEM_BLOCK_SIZE = 1000
 DEFAULT_ZOOMS_MM = (40.0, 40.0, 20.0)  # For human adults (mid-frequency), in mm
 DEFAULT_LF_ZOOMS_MM = (100.0, 100.0, 40.0)  # For human adults (low-frequency), in mm
 DEFAULT_HF_ZOOMS_MM = (16.0, 16.0, 10.0)  # For human adults (high-frequency), in mm
-BSPLINE_SUPPORT = 2 - 1.82e-3  # Disallows weights < 1e-9
 LOGGER = logging.getLogger("nipype.interface")
 
 
@@ -75,6 +74,11 @@ class _BSplineApproxInputSpec(BaseInterfaceInputSpec):
         True,
         usedefault=True,
         desc="generate a field, extrapolated outside the brain mask",
+    )
+    zooms_min = traits.Float(
+        1.4,
+        usedefault=True,
+        desc="limit minimum image zooms, set 0.0 to use the original image",
     )
 
 
@@ -123,16 +127,45 @@ class BSplineApprox(SimpleInterface):
 
     def _run_interface(self, runtime):
         from sklearn import linear_model as lm
-        from scipy.sparse import vstack as sparse_vstack
+
+        # Output name baseline
+        out_name = fname_presuffix(
+            self.inputs.in_data, suffix="_field", newpath=runtime.cwd
+        )
 
         # Load in the fieldmap
         fmapnii = nb.load(self.inputs.in_data)
-        data = fmapnii.get_fdata(dtype="float32")
-        mask = (
-            nb.load(self.inputs.in_mask).get_fdata() > 0
+        zooms = fmapnii.header.get_zooms()
+
+        # Get a mask (or define on the spot to cover the full extent)
+        masknii = (
+            nb.load(self.inputs.in_mask)
             if isdefined(self.inputs.in_mask)
-            else np.ones_like(data, dtype=bool)
+            else None
         )
+
+        need_resize = np.any(np.array(zooms) < self.inputs.zooms_min)
+        if need_resize:
+            from niworkflows.utils.images import resample_by_spacing
+
+            LOGGER.info(
+                "Resampling image with resolution exceeding 'zooms_min' "
+                f"({'x'.join(str(s) for s in zooms)})."
+            )
+            fmapnii = resample_by_spacing(fmapnii, [self.inputs.zooms_min] * 3)
+
+            if masknii is not None:
+                masknii = resample_by_spacing(masknii, [self.inputs.zooms_min] * 3)
+
+        data = fmapnii.get_fdata(dtype="float32")
+
+        # Generate a numpy array with the mask
+        mask = (
+            np.ones_like(fmapnii.dataobj, dtype=bool) if masknii is None
+            else np.asanyarray(masknii.dataobj) > 1e-4
+        )
+
+        # Convert spacings to numpy arrays
         bs_spacing = [np.array(sp, dtype="float32") for sp in self.inputs.bs_spacing]
 
         # Recenter the fieldmap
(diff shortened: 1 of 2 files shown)


oesteban:
cc/ @eilidhmacnicol

oesteban:
BTW, this is meant to be merged before the other PRs

sdcflows/interfaces/bspline.py
```diff
@@ -76,6 +75,11 @@ class _BSplineApproxInputSpec(BaseInterfaceInputSpec):
         usedefault=True,
         desc="generate a field, extrapolated outside the brain mask",
     )
+    zooms_min = traits.Float(
+        1.4,
```
effigies: I'm seeing this issue on 1.6mm and 1.8mm. I might set this as 2.0 by default, or 1.95 to be sure that we don't treat 1.98 differently from 2.0.

Would we consider a 4:1 voxel:knot ratio acceptable in the worst case? With a 16mm knot spacing and using a 2-or-nothing down-sampling ratio as proposed below, a threshold of 1.95 would get us a range of 8:1.95 to 8:1.

sdcflows/interfaces/bspline.py
```diff
@@ -123,16 +127,45 @@ class BSplineApprox(SimpleInterface):
...
+
+        need_resize = np.any(np.array(zooms) < self.inputs.zooms_min)
+        if need_resize:
+            from niworkflows.utils.images import resample_by_spacing
+
+            LOGGER.info(
+                "Resampling image with resolution exceeding 'zooms_min' "
+                f"({'x'.join(str(s) for s in zooms)})."
+            )
+            fmapnii = resample_by_spacing(fmapnii, [self.inputs.zooms_min] * 3)
```
effigies: I have [general problems](https://github.com/nipreps/niworkflows/issues/696) with resample_by_spacing. Here I think it is massive overkill. If we can agree on a safe factor-of-2 range of okay knot-to-voxel ratios, then there's a much simpler approach, where we low-pass filter and downsample without interpolation.

I found the function I was using during testing:

```Python
def _blur(in_file, target_zooms):
    import os
    import numpy as np
    import nibabel as nb
    from scipy import ndimage as ndi
    from nipype.utils.filemanip import fname_presuffix

    img = nb.load(in_file)
    factors = target_zooms[:3] / np.array(img.header.get_zooms()[:3])

    # Downsample by 2 or not at all
    scale = np.clip(factors.astype(int) // 4, 1, 2)

    if not np.any(scale > 1):
        return in_file

    out_filename = fname_presuffix(in_file, suffix='blur', newpath=os.getcwd())

    blurred = img.__class__(
        ndi.gaussian_filter(img.get_fdata(), (scale - 1) / 2),
        affine=img.affine,
        header=img.header
    )

    slicers = tuple(
        slice(1, None, 2) if factor == 2 else slice(None) for factor in scale
    )

    out = blurred.slicer[slicers]
    out.to_filename(out_filename)
    return out_filename
```

sdcflows/interfaces/bspline.py
```diff
@@ -123,16 +127,45 @@ class BSplineApprox(SimpleInterface):
...
+
+            if masknii is not None:
+                masknii = resample_by_spacing(masknii, [self.inputs.zooms_min] * 3)
```
effigies: Here we'd need to do something slightly different. Not positive if blur-downsample-(threshold at 0.5) is equivalent to a nearest-neighbor. You're also not doing a nearest-neighbor here, so it's a bug whatever downsampling method we take.

sdcflows/interfaces/bspline.py
effigies: Note that your current approach will also resample 1.2x2.0x2.0 to 1.4x1.4x1.4. The above approach would resample to 2.4x2.0x2.0.

effigies (commented):
Architectural changes look good. I think we need a different threshold and I would incline toward a simpler down-sampling method.

sdcflows/interfaces/bspline.py
```diff
@@ -457,6 +495,24 @@ def bspline_grid(img, control_zooms_mm=DEFAULT_ZOOMS_MM):
...
+    from scipy.sparse import vstack as sparse_vstack
+
+    bs_levels = []
+    weights = None
+    for sp in knot_spacing:
+        level = bspline_grid(image, control_zooms_mm=sp)
+        bs_levels.append(level)
+
+        weights = (
+            gbsw(image, level)
```
mgxd: nitpick - since we are only using this function here and space is not a concern, I feel renaming `grid_bspline_weights()` on import just obfuscates the code

sdcflows/interfaces/bspline.py
oesteban: Sounds good to me. I think 4:1 would be good in the worst case.

sdcflows/interfaces/bspline.py
oesteban: - smoothing: I don't think that's necessary - the scipy interpolation function has a `filter` option that will smooth for you.
- the problem of resolution is indeed something I'll fix. Before you even pointed at the problem, I've been talking with @eilidhmacnicol that you definitely want to be able to keep anisotropic voxels for rats. So I guess it is more about accepting either one or three floats as input.

sdcflows/interfaces/bspline.py
oesteban: Not sure I follow. This is going to interpolate at the edges of the binary mask (if binary, it doesn't have to be), so you will have some "spreadout" from the bspline kernel, but I don't think this is going to be problematic tbh.

Then, another question is whether the threshold for binarization in line 165 is too low.

effigies: I thought the mask was binary. I'll need to look again.

oesteban:
It doesn't necessarily need to be binary. I didn't respond to your concerns
about the niworkflows function. Happy to use something simpler with less
issues.

sdcflows/interfaces/bspline.py
```diff
@@ -192,6 +209,27 @@ def _run_interface(self, runtime):
...
 
+        # Interpolating in the original grid will require a new collocation matrix
+        if need_resize:
+            fmapnii = nb.load(self.inputs.in_data)
+            data = fmapnii.get_fdata(dtype="float32")
+            mask = (
+                np.ones_like(fmapnii.dataobj, dtype=bool) if masknii is None
+                else np.asanyarray(nb.load(self.inputs.in_mask).dataobj) > 1e-4
+            )
+            colmat, _ = _collocation_matrix(fmapnii, bs_spacing)
```
mgxd, line 226: this block feels redundant - we can just calculate `orig_colmat` in the first `need_resize` block and then use it if we resized

sdcflows/interfaces/bspline.py
mgxd: +1 on avoiding using the current implementation of `resample_by_spacing`

mgxd (commented):
Some notes - any hope for a few tests (with / without downsampling)?

sdcflows/interfaces/bspline.py
oesteban: this postpones the generation of a second collocation matrix to the end. The new one is only generated if the input was resized. I don't think I follow the argument here.

sdcflows/interfaces/bspline.py
oesteban: Changed to accepting tuples (i.e., anisotropic voxels).

I believe that, if we want some ratio, we should calculate the target spacings elsewhere. That looks like a good option for rodents, but I don't think we should worry about that right this minute.

sdcflows/interfaces/bspline.py
oesteban: I have implemented an alternative function. As I said before, I would leave more specialized blurring and zooms ratios for a previous interface if necessary.

sdcflows/interfaces/bspline.py
oesteban: Also, it is a large CSR matrix, so probably you want it to be as short-lived as possible.

oesteban:
LMKWYT

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,54 @@
 """Image processing tools."""
 
 
+def resample_to_zooms(in_file, zooms, order=3, prefilter=True):
```
oesteban, line 26: @effigies, does this new implementation of the resampling help with your concerns?

sdcflows/interfaces/bspline.py
```diff
@@ -124,27 +131,50 @@ class BSplineApprox(SimpleInterface):
...
+                zooms_min = [float(zooms_min)] * 3
+
+            LOGGER.info(
+                "Resampling image with resolution exceeding 'zooms_min' "
+                f"({'x'.join(str(s) for s in zooms)})."
+            )
+            fmapnii = resample_to_zooms(fmapnii, zooms_min)
+
+            if masknii is not None:
+                masknii = resample_to_zooms(masknii, zooms_min)
```
effigies: I still don't think we want to upsample in cases of anisotropic voxels. The goal is to downsample when convenient.

```suggestion
            target_zooms = np.maximum(zooms, self.inputs.zooms_min)

            LOGGER.info(
                f"Resampling image with resolution exceeding {zooms_min} "
                f"({'x'.join(str(s) for s in zooms)})."
            )
            fmapnii = resample_to_zooms(fmapnii, target_zooms)

            if masknii is not None:
                masknii = resample_to_zooms(masknii, target_zooms)
```

If we want to force isotropy, that might be a special case that deserves its own parameter.

sdcflows/interfaces/bspline.py
```diff
@@ -124,27 +131,50 @@ class BSplineApprox(SimpleInterface):
...
+
+        data = fmapnii.get_fdata(dtype="float32")
+
+        # Generate a numpy array with the mask
         mask = (
-            nb.load(self.inputs.in_mask).get_fdata() > 0
-            if isdefined(self.inputs.in_mask)
-            else np.ones_like(data, dtype=bool)
+            np.ones_like(fmapnii.dataobj, dtype=bool) if masknii is None
```
effigies: Not sure if numpy would try to coerce to an array, but we don't need this.

```suggestion
            np.ones(fmapnii.shape, dtype=bool) if masknii is None
```

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,54 @@
...
+    """Resample the input data to a new grid with the requested zooms."""
+    from pathlib import Path
+    import numpy as np
+    import nibabel as nb
+    from nibabel.affines import apply_affine
```
effigies:
```suggestion
    from nibabel.affines import rescale_affine
```

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,54 @@
...
+    affine = np.eye(4)
+    affine[:-1, :-1] = zooms * pre_affine[:-1, :-1] / pre_zooms
+
+    extremes = apply_affine(pre_affine, [(0, 0, 0), pre_shape - 1])
+    extent = np.abs(extremes[1] - extremes[0])
+
+    new_shape = np.ceil(extent / zooms).astype(int)
+    affine[:-1, -1] = apply_affine(pre_affine, (pre_shape - 1) * 0.5) - affine[
+        :-1, :-1
+    ] @ ((new_shape - 1) * 0.5)
```
effigies: How does this compare to the following?

```suggestion
    pre_zooms = np.array(in_file.header.get_zooms()[:3])
    # Could use `np.ceil` if we prefer
    new_shape = np.rint(np.array(in_file.shape[:3]) * pre_zooms / zooms)
    affine = rescale_affine(in_file.affine, in_file.shape[:3], zooms, new_shape)
```

I'd rather reuse well-tested linear algebra when we can.

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,54 @@
...
+
+    # Generate new reference
+    hdr.set_sform(affine, scode)
+    hdr.set_qform(affine, qcode)
+    newref = in_file.__class__(
+        np.zeros(new_shape, dtype=hdr.get_data_dtype()),
```
effigies: This is not a big deal, but I don't think the dtype matters, but setting it so precisely implies that it does. This might make one hesitant to change things in the future. Just flagging the thought that we might want some way to indicate what's important, like `SpatialReference(shape, affine)`, but should not hold up this PR if there isn't something that simple already in nitransforms.

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,54 @@
...
+        affine,
+        hdr,
+    )
+
+    # Resample via identity transform
+    return Affine(reference=newref).apply(in_file, order=order, prefilter=prefilter)
```
effigies, line 59: Looking at scipy.ndimage.map_coordinates, I don't think prefilter filters more or less based on the down-sampling factor.

I suspect for the factors we'll be dealing with (mostly <2, almost all <3) it's probably fine, but I don't really know how to evaluate this.

sdcflows/interfaces/bspline.py
oesteban: Not sure what you are pointing out as a problem. The PR just adds a way of imposing an upper bound to the resolution. The upper bound can be iso- or anisotropic, depending on what you pass into `zooms_min`.

What I am arguing is that, if you want specific zooms (e.g., following a 4:1 ratio), then you set up an interface to calculate them, and you pass them in. And if you want something even more specific (e.g., smoothing) then, you do the resampling outsize and set ``zooms_min = 0.0`` to ensure there's no further resampling.

sdcflows/utils/tools.py
oesteban: does rescale affine account realign the volume's center? If not, we need apply_affine to calculate the offset between extents.

sdcflows/utils/tools.py
oesteban: > Looking at scipy.ndimage.map_coordinates, I don't think prefilter filters more or less based on the down-sampling factor.

No, that's the b-spline filtering, which blurs with the width of the B-Spline basis (cubic, then 4 voxels). I believe that is enough for the purpose of this interpolation.

oesteban: In the case of SyN, ANTs is already giving you a displacements field that has been smoothed to a certain kernel width (the third parameter of -t [Syn]). We are not in a more general case where you have a signal with additive noise and you want to subsample it safely.

So I'm quite convinced no extra smoothing should be added.

sdcflows/interfaces/bspline.py
effigies: Right now, if I set the value to 1.95 and I have voxels that are 1.94x1.94x3, I will get out 1.95x1.95x1.95, which down-samples in two dimensions and up-samples in the other. If I have voxels that are 1.98x1.98x3, I get 1.98x1.98x3.

The proposal above would change it to only down-sample, so 1.94x1.94x3 would get changed to 1.95x1.95x3, which seems like less of a qualitatively different behavior when the threshold is crossed.

sdcflows/utils/tools.py
effigies: Yes. https://github.com/nipy/nibabel/blob/69307d3ac8f6127a2d1ca0e7f6d480bbf4657af0/nibabel/affines.py#L342-L379

sdcflows/interfaces/bspline.py
oesteban: Okay, I think I understand your point. Basically something like:

```Python
final_zooms = np.max(zip(old_zooms, new_zooms))
```

effigies: Yes. I had `target_zooms = np.maximum(zooms, self.inputs.zooms_min)`

```Python
>>> np.maximum((1.5, 1.5, 3), 2)
array([2., 2., 3.])
```

sdcflows/utils/tools.py
oesteban: (to answer the question, you are right, it is independent of the size ratio) -- but that should be okay, IMHO, for the reasons above.

oesteban:
I'll go ahead and merge. I plan to make a PR collecting post-merge comments, so, if something new appears, I will address it later.

sdcflows/interfaces/bspline.py
```diff
@@ -124,27 +130,48 @@ class BSplineApprox(SimpleInterface):
...
         )
-        # Create a copy of the header for use below
-        hdr = fmapnii.header.copy()
-        hdr.set_data_dtype("float32")
 
+        need_resize = np.any(np.array(zooms) < self.inputs.zooms_min)
+        if need_resize:
+            from sdcflows.utils.tools import resample_to_zooms
+
+            zooms_min = np.maximum(zooms, self.inputs.zooms_min)
```
effigies, line 154: Just for clarity, I think we want to rename from `zooms_min`, since it now diverges form `self.inputs.zooms_min`.

oesteban: what would you suggest? (I have no particular preference)

effigies: `target_zooms` made sense to me.

PR #302, opened 2022-11-17
MAINT: Housekeeping and more verbose debugging outputs

oesteban:


sdcflows/transform.py
```diff
@@ -413,12 +426,33 @@ def grid_bspline_weights(target_nii, ctrl_nii):
...
-    newaff = xfm.matrix @ coeff.affine
-    return coeff.__class__(coeff.dataobj, newaff, coeff.header)
+    hdr = coeff.header.copy()
+
+    if fmap_target is not None:  # Debug mode: generate fieldmap reference
+        nii_target = nb.load(fmap_target)
+        debug_ref = (~xfm).apply(fmap_ref, reference=nii_target)
+        debug_ref.header.set_qform(nii_target.affine, code=1)
+        debug_ref.header.set_sform(nii_target.affine, code=1)
+        debug_ref.to_filename(Path() / "debug_fmapref.nii.gz")
```
effigies, line 443: Odd idiom.

```Python
>>> Path() / "debug_fmapref.nii.gz"
PosixPath('debug_fmapref.nii.gz')
```

So this is the same as:

```Python
        debug_ref.to_filename("debug_fmapref.nii.gz")
```

sdcflows/transform.py
```diff
@@ -413,12 +426,33 @@ def grid_bspline_weights(target_nii, ctrl_nii):
...
+
+    # Generate a new transform
+    newaff = np.linalg.inv(np.linalg.inv(coeff.affine) @ (~xfm).matrix)
```
effigies, line 446: I don't understand why you're replacing `A @ B` with `~(~B @ ~A)`.

sdcflows/workflows/apply/correction.py
```diff
@@ -103,7 +103,9 @@ def init_unwarp_wf(omp_nthreads=1, debug=False, name="unwarp_wf"):
 
     rotime = pe.Node(GetReadoutTime(), name="rotime")
     rotime.interface._always_run = debug
-    resample = pe.Node(ApplyCoeffsField(num_threads=omp_nthreads), name="resample")
+    resample = pe.Node(ApplyCoeffsField(
+        num_threads=omp_nthreads if not debug else 1
```
effigies, line 107: Just as a thought, I wonder if this would be more or less clear:

```Python
    resample = pe.Node(ApplyCoeffsField(num_threads=debug or omp_nthreads), name="resample")
```

Since bools are a subclass of int, it will be 1 if debug is True or omp_nthreads if debug is False.

```Python
>>> ApplyCoeffsField(num_threads=True).inputs

in_coeff = <undefined>
in_data = <undefined>
in_xfms = <undefined>
num_threads = 1
pe_dir = <undefined>
ro_time = <undefined>
```

effigies (commented):
Small comments. Nothing that would have been blocking.

effigies:
Can be backported to 2.1.0 if needed.

PR #303, opened 2022-11-17
FIX: Do not reorient distorted image in apply

oesteban:
Ensuring positive cosines should not be necessary.


sdcflows/interfaces/bspline.py
```diff
@@ -581,19 +579,17 @@ def _b0_resampler(in_file, coeffs, pe, ro, hmc_xfm=None, unwarp=None, newpath=No
...
     else:
         retval[2] = None
 
-    # Unwarp the reoriented image, and restore original orientation
-    unwarped_img = reorient_image(
-        unwarp.apply(reoriented_img, ro_time=ro, pe_dir=pe),
-        axcodes,
-    )
+    # Unwarp
+    unwarped_img = unwarp.apply(distorted_img, ro_time=ro, pe_dir=pe)
```
effigies, line 641: Was the idea to have the corrected image in the same space as the displacement field? If so, would we still want to reorient after unwarping?

oesteban: I'm decomposing #294 into independent patches, and I had this in it. I still need to understand why the test is failing. Since we are moving the coefficients into the distorted image space (without reorientations), here it should be possible to apply the transform directly (and avoid reorients completely).

I hope I find some time next week to think/resolve the failing test and give you a better answer.

oesteban: Okay, I have figured it out.

- Your question: thanks to this PR, the original image, the field, and the corrected image are all in the same space, without the need for reorientations along the way.

- Further scope: thanks to eliminating this reorientation, I have caught a bug in our tests. The "fixed" topup coefficients (i.e., with a proper affine) did not have the internal LR flip of FSL. The example has been fixed, and now all tests pass locally.

PR #304, opened 2022-11-17
ENH: Evaluate B-Splines using scipy

oesteban:
As suggested by @effiges, it was indeed doable (1D, and I believe it can do 2D, not sure about 3D).

This implementation is slower, but removes maintenance from ourselves. The difference in runtime is negligible (i.e., maybe 2 secs slower in 20 sec).


sdcflows/transform.py
```diff
@@ -406,24 +395,20 @@ def grid_bspline_weights(target_nii, ctrl_nii):
...
         zip(origin, shape, target_nii.header.get_zooms()[:3])
     ):
-        locations = np.arange(0, n, dtype="float16") * sp / ctrl_sp[i] + o
-        knots = np.arange(0, ctrl_nii.shape[i], dtype="float16")
-        distance = np.abs(locations[np.newaxis, ...] - knots[..., np.newaxis])
-
-        within_support = distance < 2.0
-        d_vals, d_idxs = np.unique(distance[within_support], return_inverse=True)
-        # Calculate univariate B-Spline weight for samples within kernel spread
-        weights_vals = _cubic_bspline(d_vals)[d_idxs].astype("float32")
```
effigies, line 416: I think my thought was to use [scipy.signal.cubic](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.cubic.html) (a special case of [scipy.signal.bspline](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.bspline.html)):

```Python
weights_vals = cubic(d_vals)[d_idxs].astype("float32")
```

oesteban: as per the documentation I think that would be exactly equivalent to the `np.piecewise` that we had before. But now I see what you meant.

From an implementation perspective, I think the design route is more consistent with the rest of scipy so perhaps makes the code more readable.

effigies: Sounds good. Just aiming for clarity.

sdcflows/transform.py
```diff
@@ -397,33 +386,37 @@ def grid_bspline_weights(target_nii, ctrl_nii):
...
+
+        # Calculate the index component of samples w.r.t. B-Spline knots along current axis
+        x = nb.affines.apply_affine(target_to_grid, coords.T)[:, axis]
+
+        # BSpline.design_matrix requires all x be within -4 and 4 padding
+        # This padding results from the B-Spline degree (3) plus one
+        t = np.arange(-4, knots_shape[axis] + 4, dtype=dtype)
+
+        # Calculate K x N collocation matrix (discarding extra padding)
+        colloc_ax = BSpline.design_matrix(x, t, 3)[:, 2:-2]
```
effigies, line 415: Aren't we adding four on each side? Shouldn't this be:

```suggestion
        colloc_ax = BSpline.design_matrix(x, t, 3)[:, 4:-4]
```

oesteban: This is confusing. `BSpline.design_matrix` already removes 2 knots at each end.

If I don't pad the knots `t` with 4 (instead of just 2), then we get out-of-bounds errors.

I have checked with our former implementation and doing this particular padding (4 at beginning and end) and de-padding (2, beginning and end), the weights we obtain with this and the former method are very close (rtol ~ 1e-4).

effigies: Okay, sounds good.

PR #305, opened 2022-11-24
FIX: Calculate collocation matrix with rotated coefficients

### LINKED ISSUE #218: Unexplained flip of the PE direction under some unknown circumstances

### LINKED ISSUE #289: B-Spline grid and target EPI are assumed to have the same direction cosines

oesteban:
A cythonized implementation of the tensor product B-Splines is proposed.

Resolves: https://github.com/nipreps/sdcflows/issues/289.
Resolves: https://github.com/nipreps/sdcflows/issues/218.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -98,8 +98,8 @@ jobs:
       timeout-minutes: 5
       run: |
         $CONDA/bin/pip install -r min-requirements.txt
-        $CONDA/bin/pip install .[tests]
-
+        $CONDA/bin/python setup.py build_ext --inplace
+        $CONDA/bin/pip install -e .[tests]
 
     - uses: actions/cache@v2
       with:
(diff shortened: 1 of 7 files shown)


oesteban:
I was making good progress on this one until I went to BrainHack Donostia two weeks ago, and couldn't pay a lot of attention to it. However, serendipity put @fabianp in the same room, and he gave me some advice to speed up the B-Spline approximation (thanks, Fabian!). I will explain it later in our meeting and try to figure out some timeline to finish this.

Issue #306, opened 2022-12-01
Limits on rotations when combining EPIs?


effigies:
Inspired while thinking about #297: for SyN-SDC in fMRIPrep, we collate BOLD series based on their effective echo spacing/total readout time and phase-encoding direction. If you consider cases where subjects adjust their heads between runs or leave the scanner, it seems like there is potential to have significantly different distortion profiles.

Consider a case where in session 1, the subject's head is rotated ~15deg to the left of plumb about the vertical axis (in the axial plane), and ~15deg to the right in session 2. You would now have a 30deg difference in the direction of susceptibility artifacts between the two scans. An average EPI reference would increase noise, rather than reduce it. Rotations about the transverse axis will also impact the distortion.

If this reasoning is correct, it's possible that this is responsible for some SyN-SDC artifacts in multi-run datasets. I will be testing a multi-run dataset on a per-run basis (via the expedient of just changing the TRT because it doesn't matter).

If this is the case, perhaps the "correct" solution is to estimate the fieldmap in T1 space and average that across BOLD images, rather than averaging the BOLD references and estimating the fieldmap. A compromise for efficiency could be to estimate the alignment between BOLD references and cluster images that are sufficiently close together.

effigies:
Per meeting, try to get wrangler to generate 1 fieldmap per BOLD. If this works, get it into 2.2.0

PR #312, opened 2022-12-06
ENH: Split SyN fieldmap estimates per-EPI

effigies:
This is an experimental PR to undo the merging of individual EPI images which may not be well-aligned with respect to the phase-encoding direction. This will revert us to the basic collation method of fMRIPrep 20.2.x while preserving the fit-apply split introduced in fMRIPrep 21.0.0.

This experiment is worth performing before continuing to optimize b-spline approximation of the SyN-estimated field.

I've tried to write this with an eye to making it easier to plug in grouping strategies in the future.

First step in fully addressing #306.

### DIFF
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -499,28 +499,23 @@ def find_estimators(
             meta.update({"TotalReadoutTime": ro})
             epi_targets.append(fm.FieldmapFile(candidate.path, metadata=meta))
 
-        for pe_dir in sorted(set(pe_dirs)):
-            pe_ro = [ro for ro, pe in zip(ro_totals, pe_dirs) if pe == pe_dir]
-            for ro_time in sorted(set(pe_ro)):
-                fmfiles, fmpaths = tuple(
-                    zip(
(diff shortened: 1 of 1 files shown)


effigies:
When splitting, I get normal looking fields (only one shown, as they all look alike):

image

But the conversion to the individual subject space varies across images (I see two like each):

image
image

Unsurprisingly distortion correction looks much more reasonable when the run-space fieldmaps look reasonable (though neither is as bad as some failures we've seen):

Screenshot from 2022-12-07 08-17-39
Screenshot from 2022-12-07 08-18-21

What's odd is that all the input images have the same affine and the only difference I'm seeing in metadata is a few of the slice times.

effigies:
Oh man. There's a collation problem here, where the low-frequency maps are associated with odd runs and high-frequency with even.

effigies:
Addressed in #317.

effigies:
This does not seem to be necessary for the test cases I'm looking at. I suggest we hold off until 23.0 to see whether problems are resolved for most users with our other updates. Then we could make this an option.

PR #314, opened 2022-12-07
ENH: Default to 4mm re-zoom for b-spline approximation

effigies:
Downsample fieldmap images to 4mm isotropic before approximating b-splines.

The 1.95mm suggestion was a cut-off for downsampling by a factor of 2 so that we did not go worse than 4mm. (The 0.05 margin was to avoid something like 1.99x1.99x2.01 from becoming 3.98x3.98x2.01.) Given that instead we're resampling to a maximum resolution, let's just set it

Running at 2mm, I'm still finding multi-hour approximations. 4mm seems reasonable, as that's still a 4:1 voxels-to-knots ratio for the high-frequency components. I haven't tried 3mm.


effigies:
This was discussed in meeting. Merging.

PR #315, opened 2022-12-07
ENH: Drop n_procs tag from BSplineApprox

effigies:
BSplineApprox does not make use of multiple processors by default or accept them as an option.


effigies:
Going to merge, but please let me know if you object before a release, @oesteban. It may be that it can go multithread in cases I don't understand.

PR #319, opened 2022-12-09
FIX: Pad BSpline design matrix

effigies:
If the bounding box of the target image goes outside the bounding box of the B-spline coefficient image, scipy will refuse to construct a design matrix. This is because the coordinates `x` (voxel indices translated into non-integral indices in the coefficient image's voxel space) may go below 0 or above the number of knots.

This PR aims to work around this limitation without changing the logic on the grounds that we need the best approximation of the field at all locations within the target image. At the edges, the field is likely to get masked out anyway. This assumes that the fieldmap and target image are in register and the differences in their affines reflect real differences in FoV.

We have a default padding of 4 knots in each direction, which gets reduced to 2 by `BSpline.design_matrix`. This adds in a padding based on the min/max values of `x`. The padding is added to `t` and then removed from the design matrix to ensure that the weights still correspond to the coefficients stored in the coefficient images.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -406,13 +406,15 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
 
         # Calculate the index component of samples w.r.t. B-Spline knots along current axis
         x = nb.affines.apply_affine(target_to_grid, coords.T)[:, axis]
+        pad_left = max(int(-np.rint(x.min())), 0)
+        pad_right = max(int(np.rint(x.max()) - knots_shape[axis]), 0)
 
         # BSpline.design_matrix requires all x be within -4 and 4 padding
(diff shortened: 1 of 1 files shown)


effigies:
@mgxd The result isn't great, but at least it runs:

image

effigies:
I think the issue may be the translation:

image
image

I'm going to see what happens if I simply shift `t` by `x.min()`.

effigies:
> I'm going to see what happens if I simply shift `t` by `x.min()`.

That made it worse. Is this at least back to where it was previously?

mgxd:
Yup - I'm okay with merging this so we can test with https://github.com/nipreps/sdcflows/pull/317

effigies:
I tested with both to get the above. But let's merge this at least.

PR #321, opened 2022-12-12
ENH: Dynamically choose number of resampling threads to adapt to memory constraints

effigies:
This PR by default makes `unwarp_wf.resample` correctly estimate its memory usage. If passed a `free_mem` variable, such as `fmriprep.config.execution.free_mem`, then it can reduce the number of threads to adapt to the available memory.

### DIFF
--- a/sdcflows/workflows/apply/correction.py
+++ b/sdcflows/workflows/apply/correction.py
@@ -26,7 +26,7 @@ from nipype.interfaces import utility as niu
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 
 
-def init_unwarp_wf(omp_nthreads=1, debug=False, name="unwarp_wf"):
+def init_unwarp_wf(free_mem=None, omp_nthreads=1, debug=False, name="unwarp_wf"):
     r"""
     Set up a workflow that unwarps the input :abbr:`EPI (echo-planar imaging)` dataset.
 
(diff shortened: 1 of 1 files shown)


effigies:
@oesteban @mgxd I plan to merge and release ASAP unless one of you objects, so we can get fMRIPrep out today with this patch in.

Please review if you have concerns.

oesteban:
This makes the decision of resampling dependent on a situational parameter very specific to the platform (e.g., we might expect very different behavior on the cloud, clusters, and desktops). For run-to-run replication, we will need to keep this decision stored somewhere next to, e.g., random seeds.

In addition, I don't think the subsampling is particularly dangerous in this case - so I would prefer not to add the decision and resample if the resolution is high enough.

That said, this is not a hill I'm willing to die on.

effigies:
@oesteban This should have no impact on the actual resampling, just how many threads we run in parallel. I'm not sure if you misread what I proposed or there's a non-determinism here that I'm not seeing.

oesteban:
Apologies, I totally misunderstood the goal of the PR. I'm going to look at the code more carefully.

sdcflows/workflows/apply/correction.py
```diff
@@ -26,7 +26,7 @@
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 
 
-def init_unwarp_wf(omp_nthreads=1, debug=False, name="unwarp_wf"):
+def init_unwarp_wf(free_mem=None, omp_nthreads=1, debug=False, name="unwarp_wf"):
```
mgxd: just as a nitpick - since this function isn't keyword only, this could throw off any code using this positionally
```suggestion
def init_unwarp_wf(*, free_mem=None, omp_nthreads=1, debug=False, name="unwarp_wf"):
```

mgxd (approved):
LGTM

oesteban (approved):
Looks great thanks!

effigies:
I've tried to update the title to make a better changelog entry. Please feel free to rename if that still doesn't capture it.

PR #322, opened 2022-12-17
CI: Switch to miniconda setup, install fsl tools through conda

effigies:
Seeking a longer-term solution than #310.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
 
     steps:
     - uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


effigies:
@mgxd @oesteban Any objection? Should we pin to specific versions of FSL tools? We could add a bleeding-edge test in addition to release-pinned tests, so we can report breakage upstream.

oesteban (approved):
LGTM!

.github/workflows/unittests.yml
```diff
@@ -34,15 +34,15 @@ jobs:
     strategy:
       max-parallel: 5
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9]
```
mgxd, line 37: worth adding 3.10?
```suggestion
        python-version: ["3.8", "3.9", "3.10"]
```

mgxd (approved):
lgtm

mgxd:
ah, too slow!

oesteban:
Does the fsl channel have a candidate for 3.10? That could be a future culprit

effigies:
The FSL tools we install shouldn't interact with Python versions. We can test on 3.10 in a new PR.

PR #327, opened 2023-01-05
CI: Update docker/machine images for 1.3.x branch

effigies:
This PR does the following:

1) Gets CircleCI builds working on `maint/1.3.x`
    * Upgrades machines/docker images and adjusts pyenv/`python -m`/sudo usage as needed
    * Updates neurodebian configuration
1) Retrieves environment variables from context (backporting #329)
1) Cordons off `master` and `maint/1.3.x` Docker builds
    * Caches now look for `build-{version}-maint/1.3.x` instead of `master`, and do not fall back to `build-{version}-`
    * Fetches Docker image corresponding to latest git tag, rather than `master`, to ensure more consistent builds within a minor release series
1) Updates the docs to more closely match master. This may not handle versioning correctly, but it does build. We can try again if we do have another release.
1) Upgrades pip in the Dockerfile before installing the package; pip's resolver changes over time and the old ones couldn't handle the transitive upgrade sdcflows -> nilearn -> numpy, so we got numpy 1.15.x while nilearn required more recent.

Another shot at #222. Starting just by updating the images.


effigies:
@mgxd @oesteban This is not urgent, as there are no unreleased fixes on the maintenance branch, but I think would be worth getting in when you have a chance to review. It's grown somewhat large, so hopefully the above comment will guide you.

There are a couple things worth merging forward to `master`, namely the `pip install --upgrade pip` in docker and the retrieval of docker images based on the latest git tag.

effigies:
@mgxd Would you mind reviewing when you get a chance? (This isn't blocking, just going through outstanding PRs and nudging reviewers if they're ready.)

PR #334, opened 2023-01-31
ENH: Calculate fieldwarps in reference space in unwarp_wf

### LINKED ISSUE #333: Generate untransformed fieldmap in reference space
`unwarp_wf` currently returns a list of transforms calculated in each BOLD volume's space (calculated by the inverse of the HMC transform to reference space). However, it is not always the case that these transforms can be used in later steps, so one possible action is to take the first transform. That transform is not in reference space, but instead in the first volume's space, which can potentially differ significantly if sbrefs are used.

It should be possible (and reasonably cheap) to calculate an SDC transform with an identity motion correction transform to ensure that the HMC (vol -> boldref) + SDC (boldref -> boldref) chain in T1 and template transformations is correct.

This should not affect BOLD native-space outputs.

I assume this all applies with minor modifications to DWI workflows.

effigies:
Fixes #333.

This is more a bug in fmriprep than in sdcflows proper.

### DIFF
--- a/sdcflows/workflows/apply/correction.py
+++ b/sdcflows/workflows/apply/correction.py
@@ -51,6 +51,8 @@ def init_unwarp_wf(*, free_mem=None, omp_nthreads=1, debug=False, name="unwarp_w
     ------
     distorted
         the target EPI image.
+    distorted_ref
+        the distorted reference EPI image to which each volume has been motion corrected
     metadata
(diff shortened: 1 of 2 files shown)


mgxd (approved):
In theory this sounds reasonable to me

effigies:
Cool. I guess we need to figure out what's going on with CI.

PR #335, opened 2023-02-23
Correcting the condition to perform the deoblique

yasseraleman:
The condition was always TRUE even if the values were below the specified threshold


effigies:
If I'm reading it right, any obliquity besides 0.0 was triggering the `False` condition and going on to be deobliqued. So this was effectively an over-stringent test. I agree with the fix; not sure if its impact will be very large. Do you have a test dataset where you can demonstrate the effect of this change?

effigies:
@yasseraleman I will merge this without a test if needed, but if you have a demonstration of where this causes problems, I would appreciate it.

PR #337, opened 2023-03-01
TEST: Squeeze image before passing to SimpleBeforeAfter

effigies:
Trying to fix the CI failures by reproducing locally. First problem is I can't build the image.

After reproduction, the issue turned out to be we were passing a 4D image to SimpleBeforeAfter, and we just need to squeeze it.

Includes #330, since I needed that to test locally.

Closes #330.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,8 +131,8 @@ jobs:
 
       - restore_cache:
           keys:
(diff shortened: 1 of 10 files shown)


effigies:
This is necessary for a release but with no significant code changes except a patch for tests that only run if reporting outputs are to be generated. I'm going to merge without review.

PR #338, opened 2023-03-01
REL: 2.3.0

effigies:
Release prep for 2.3.0.

I think it should probably be 2.3.0 because of #334, but strictly that adds nodes to a workflow, but does not change any existing connections. This could be 2.2.3, if people would rather not crank the version and possibly need to maintain `maint/2.2.x` separately.

Ping @mgxd and @oesteban for comments.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,15 @@
+2.3.0 (March 01, 2023)
+======================
+New feature release in the 2.3.x series.
+
+This release supports fMRIPrep 23.0.x and Nibabies 23.0.x.
+
+* ENH: Calculate fieldwarps in reference space in unwarp_wf (#334)
+* TEST: Squeeze image before passing to SimpleBeforeAfter (#337)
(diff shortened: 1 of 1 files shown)


oesteban:
I vote for the easiest to maintain, which looks like 2.3.0.

Even if for whatever reason we decide we need to cut 2.4.0 next, it would be easy to recall 2.3.0 to avoid a maintenance branch with just one feature.

CHANGES.rst
```diff
@@ -1,3 +1,15 @@
+2.3.0 (March 01, 2023)
+======================
+New feature release in the 2.3.x series.
+
+This release supports fMRIPrep 23.0.x and Nibabies 23.0.x.
```
effigies, line 5: @mgxd Is this a correct line from the nibabies perspective?

mgxd: We've already released 23.0.0, but whether next release is 23.0.1 or 23.1.0, and because we talked about lessening version restrictions, this release can be incorporated

effigies: K.

PR #342, opened 2023-03-20
RF: Split PEPolar fieldmaps by intent, if available

effigies:
Encountered a dataset (cc @demidenm) with 6 fieldmaps, identifiable with `IntendedFor`. SDCFlows combined them into a two fieldmaps with three scans in each direction.

I understand that one goal was to provide better, more uniform fieldmaps, but I don't think that should come at the cost of ignoring clear intent. If the user wants to combine the fieldmaps to use on all images, they may do so explicitly.

This approach also groups fieldmaps with missing `IntendedFor` or with an empty list as a value. It would be the responsibility of downstream code to interpret whether to apply these fieldmaps to any EPI data.

PR #343, opened 2023-03-20
ENH: Consistently log failures to form fieldmaps

effigies:
Follow up to #341. Initially proposed as part of #342, but these improvements are non-breaking, so might as well get them in and released.


effigies:
@mgxd

mgxd (approved):
LGTM, much needed when attempting to debug not in a dev environment.

Side note on `find_estimators()` as a whole: we should probably refactor this function down the line, as it has become a bit of a rats nest

Issue #336, opened 2023-02-23
Support BIDS-URIs in `IntendedFor`


effigies:
### What would you like to see added in this software?

In an effort to unify the disparate ways of encoding within-dataset paths and remove ambiguity, BIDS-URIs were introduced. You can now end up with `IntendedFor = ['bids::sub-01/func/sub-01_task-rest_bold.nii.gz']` where previously you would have had `IntendedFor = ['func/sub-01_task-rest_bold.nii.gz']`.

We should resolve BIDS-URIs in the fieldmap wrangler. Erroring if we cross subject lines is fine.

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

See https://github.com/nipreps/fmriprep/issues/2947

PR #349, opened 2023-04-17
ENH: Resolve BIDS-URIs

effigies:
First attempt at #336. Will need to do some integration testing.


oesteban (dismissed):

oesteban (commented):
fix: import future annotations

oesteban (approved):

PR #354, opened 2023-04-20
FIX: Reorient fieldmaps to RAS before estimating B-splines

### LINKED ISSUE #353: Distortion correction translating instead of unwarping
### What happened?

I am running fMRIPrep 23.0.1 on a dataset with two magnitudes and one phasediff in fmap/. From the svg figure in the visual report, it seems that SDCflows did translation only. I have compared the boldref file in the native space with the unprocessed bold file and the output indeed looks the same as the unprocessed bold except for the translation. 
I have also used 23.0.1 with other datasets with phasediff or direct fieldmapping, but those were unwarped properly with noticeable changes. The only difference I can spot so far is that the PE direction in the problematic dataset is Right-Left while the PE direction in other datasets is Anterior-Posterior. So I am thinking the problem might be with handling left-right/right-left EPI and/or orientation.
I have attached the distorted, corrected and the fieldmap preview from the visual report.
Please let me know if you need any additional information. Thank you.

### What command did you use?

```shell
docker run -it --rm -v $PWD/fortesting:/data:ro -v $PWD/fortesting_out:/out -v /opt/freesurfer/license.txt:/opt/freesurfer/license.txt nipreps/fmriprep:23.0.1 /data /out participant --fs-no-reconall --md-only-boilerplate --output-spaces T1w
```

### What version of fMRIPrep are you running?

23.0.1

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

No

```shell
No errors to report
```

distorted
corrected
fieldmap

effigies:
In #339, we ensured that the b-spline coefficients were interpreted in RAS when applied, but failed to ensure that they were calculated in RAS when deriving from pre-existing fieldmaps.

I still need to carefully walk through a workflow directory created fresh with this patch to verify that everything is as it should be.

Fixes #353.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -181,17 +181,20 @@ class BSplineApprox(SimpleInterface):
         )
 
         # Recenter the fieldmap
+        center = 0
         if self.inputs.recenter == "mode":
             from scipy.stats import mode
 
             # Handle pre- and post-1.9 mode behavior.
             # squeeze can be dropped when the minimum version reaches 1.9
             # Will become: data -= mode(data[mask], keepdims=False).mode
-            data -= np.squeeze(mode(data[mask]).mode)
+            center = np.squeeze(mode(data[mask]).mode)
         elif self.inputs.recenter == "median":
-            data -= np.median(data[mask])
+            center = np.median(data[mask])
         elif self.inputs.recenter == "mean":
-            data -= np.mean(data[mask])
+            center = np.mean(data[mask])
+
+        data -= center
 
         # Calculate collocation matrix from (possibly resized) image and knot grids
         colmat = sparse_vstack(grid_bspline_weights(fmapnii, grid) for grid in bs_grids).T.tocsr()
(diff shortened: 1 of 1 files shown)


effigies:
Okay. I had to be a bit more thorough to ensure that the debug maps were generated properly, but this looks good now. With this, the outputs of `bs_filter` are always in RAS. So I think we have:

1) PEPolar: Orig -> ReorientToLAS -> LAS -> TOPUP -> LAS (affine says RAS) -> TOPUPCoeffReorient -> RAS
2) phasediff/phase1+2: Orig -> Phasediff2Fieldmap -> Orig -> BSplineApprox -> RAS
3) Fieldmap: Orig -> CheckB0Units -> Orig -> BSplineApprox -> RAS
4) SyN: Orig -> SyN -> LPS (affine says RAS) -> DisplacementsField2Fieldmap -> RAS -> BSplineApprox -> RAS

effigies:
@oesteban Do you have any time to review? I consider this a bug fix and have tested locally and I'm okay merging if tests are green.

PR #355, opened 2023-04-27
ENH: Function to calculate reference grids aligned with the coefficients

oesteban:
This is split off from #346, to focus only on this particular task, which is probably the most complicated aspect of the refactor (or the one we definitely want to get right).

### DIFF
--- a/sdcflows/utils/tests/test_tools.py
+++ b/sdcflows/utils/tests/test_tools.py
@@ -22,8 +22,10 @@
 #
 """Test EPI manipulation routines."""
 import numpy as np
+import pytest
 import nibabel as nb
-from ..tools import brain_masker
+from nitransforms.linear import Affine
+from sdcflows.utils.tools import brain_masker, deoblique_and_zooms
 
 
 def test_epi_mask(tmpdir, testdata_dir):
@@ -31,3 +33,36 @@ def test_epi_mask(tmpdir, testdata_dir):
     tmpdir.chdir()
     mask = brain_masker(testdata_dir / "epi.nii.gz")[-1]
     assert abs(np.asanyarray(nb.load(mask).dataobj).sum() - 166476) < 10
+
+
+@pytest.mark.parametrize("padding", [0, 1, 4])
+@pytest.mark.parametrize("factor", [1, 4, 0.5])
+def test_deoblique_and_zooms(padding, factor):
+    """Check deoblique and denser."""
+
+    # Generate an example reference image
+    ref_data = np.zeros((20, 30, 40), dtype=np.float32)
+    ref_affine = np.eye(4)
+    ref_affine[:3, :3] = np.diag([1.0, 1.2, 0.8])  # Set zooms to (2, 3, 4)
+    ref_img = nb.Nifti1Image(ref_data, ref_affine)
+    ref_zooms = np.array(ref_img.header.get_zooms()[:3])
+
+    # Generate an example oblique image
+    ob_data = np.ones_like(ref_data)
+    rotate = np.eye(4)
+
+    # Rotate 90 degrees around x-axis
+    rotate[:3, :3] = np.array([[0, 1, 0], [0, 0, 1], [1, 0, 0]])
+    ob_img = nb.Nifti1Image(ob_data, rotate @ ref_affine)
+
+    # Call function with default parameters
+    out_img = deoblique_and_zooms(ref_img, ob_img, padding=padding, factor=factor)
+
+    # Check output shape and zooms
+    assert np.allclose(out_img.header.get_zooms()[:3], ref_zooms / factor)
+
+    resampled = Affine(reference=out_img).apply(ob_img, order=0)
+    ref_volume = ref_data.size * ref_zooms.prod()
+    res_volume = resampled.get_fdata().sum() * np.prod(resampled.header.get_zooms())
+    # 20% of change in volume is too high, must be an error somewhere
+    assert abs(ref_volume - res_volume) < ref_volume * 0.2
--- a/sdcflows/utils/tests/test_tools.py
+++ b/sdcflows/utils/tests/test_tools.py
@@ -36,22 +36,30 @@ def test_epi_mask(tmpdir, testdata_dir):
 
 
 @pytest.mark.parametrize("padding", [0, 1, 4])
-@pytest.mark.parametrize("factor", [1, 4, 0.5])
-def test_deoblique_and_zooms(padding, factor):
+@pytest.mark.parametrize("factor", [1, 4, 0.8])
+@pytest.mark.parametrize("centered", [True, False])
+@pytest.mark.parametrize("rotate", [
+    np.eye(4),
+    # Rotate 90 degrees around x-axis
+    np.array([[0, 1, 0, 0], [0, 0, 1, 0], [1, 0, 0, 0], [0, 0, 0, 1]]),
+])
+def test_deoblique_and_zooms(tmpdir, padding, factor, centered, rotate):
     """Check deoblique and denser."""
+    tmpdir.chdir()
 
     # Generate an example reference image
-    ref_data = np.zeros((20, 30, 40), dtype=np.float32)
-    ref_affine = np.diag([1.0, 1.2, 0.8, 1.0])
+    ref_data = np.zeros((20, 32, 40), dtype=np.float32)
+    ref_data[1:-2, 10:-11, 1:-2] = 1
+    ref_affine = np.diag([2.0, 1.25, 1.0, 1.0])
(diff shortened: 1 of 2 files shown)


oesteban:
A test should come shortly - cleaning up some of the code of ChatGPT :D

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,71 @@
...
+        oblique.affine,
+        np.array(list(product((0, 1), repeat=3))) * (ref_shape - 1),
+    )
+    extent_ijk = apply_affine(np.linalg.inv(affine), extent)
+
+    if isinstance(padding, int):
+        padding = tuple([padding] * 3)
+
+    underflow = min(extent_ijk.min(0) - padding, 0)
+    overflow = max(extent_ijk.max(0) + padding + 1 - ref_shape, 0)
```
effigies: This doesn't make sense to me. I think you probably want to be using `np.min`/`np.max` or `np.minimum`/`np.maximum`.

sdcflows/utils/tests/test_tools.py
```diff
@@ -23,11 +23,50 @@
...
     assert abs(np.asanyarray(nb.load(mask).dataobj).sum() - 166476) < 10
+
+
+def test_deoblique_and_zooms():
+    """Check deoblique and denser."""
+
+    # Generate an example reference image
+    ref_data = np.zeros((20, 30, 40), dtype=np.float32)
+    ref_affine = np.eye(4)
+    ref_affine[:3, :3] = np.diag([2, 3, 4])  # Set zooms to (2, 3, 4)
```
effigies:
```suggestion
    ref_affine = np.diag([2, 3, 4, 1])  # Set zooms to (2, 3, 4)
```

sdcflows/utils/tests/test_tools.py
```diff
@@ -23,11 +23,50 @@
...
+    ref_img = nb.Nifti1Image(ref_data, ref_affine)
+
+    # Generate an example oblique image
+    ob_data = np.zeros((25, 35, 45), dtype=np.float32)
+    ob_affine = np.eye(4)
+
+    # Rotate 90 degrees around x-axis
```
effigies: This test seems fine to test, but I would also add a 30deg test.

sdcflows/utils/tests/test_tools.py
```diff
@@ -23,11 +23,50 @@
...
+    ob_affine[:3, :3] = np.array([[0, 1, 0], [0, 0, 1], [1, 0, 0]])
+    ob_img = nb.Nifti1Image(ob_data, ob_affine)
+
+    # Call function with default parameters
+    out_img = deoblique_and_zooms(ref_img, ob_img)
+
+    # Check output shape and zooms
+    assert out_img.shape == (26, 36, 46)
+    assert np.allclose(out_img.header.get_zooms()[:3], (0.5, 1.0, 1.0))
```
effigies: Some explanation would help. Why are the zooms 0.5x1x1? If we're upsampling by a factor of 4, shouldn't it be 0.5x0.75x1.0? And if we're padding by 1 in each direction, shouldn't the shape be (27, 37, 47)?

And what impact does the 90deg rotation have? Should it be 0.5x1.0x0.75 and (27, 47, 37)?

sdcflows/utils/tests/test_tools.py
oesteban: Yes, I was too lazy to pre-calculate what the shapes must result for several padding/factor combinations so opted for looking at the volume after interpolation. I have to check why there seems to be a substantial error when doing that. Once I figure that out, a 30 deg test (or whatever) could be added with parametrization (and tested with factors/paddings)

sdcflows/utils/tests/test_tools.py
oesteban: Sorry, I was still on it. I can give you a heads up when the test is ready.

oesteban:
While debugging this, I've found:

- nipy/nitransforms#175
- Mango does not properly interpret x-forms with large zoom factors, apparently. Images looked good on freeview though.

oesteban:
@effigies this is now ready. I submitted it without the draft marking because I thought the initial version was more mature and testing was going to be easier, which didn't happen.

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,70 @@
...
+    ref_zooms = np.array(hdr.get_zooms()[:3])
+    _, scode = in_reference.get_sform(coded=True)
+    _, qcode = in_reference.get_qform(coded=True)
+
+    # Calculate the 8 most extreme coordinates of oblique in in_reference space
+    extent = apply_affine(
+        oblique.affine,
+        np.array(list(product((0, 1), repeat=3))) * (np.array(oblique.shape[:3]) - 1),
+    )
+    extent_ijk = apply_affine(np.linalg.inv(affine), extent)
```
effigies: I think it's clearer to give the complicated expression a name. We can keep it to two statements by composing the affines:

```suggestion
    corners = np.array(list(product((0, 1), repeat=3))) * (np.array(oblique.shape[:3]) - 1)
    extent_ijk = apply_affine(np.linalg.inv(affine) @ oblique.affine, corners)
```

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,70 @@
...
+
+    underflow = np.clip(extent_ijk.min(0) - padding, None, 0).astype(int)
+    overflow = np.ceil(
+        np.clip(extent_ijk.max(0) + padding + 1 - ref_shape, 0, None)
+    ).astype(int)
+    if np.any(underflow < 0) or np.any(overflow > 0):
+        # Add under/overflow voxels
+        ref_shape += np.abs(underflow) + overflow
```
effigies: Underflow is always non-positive.

```suggestion
        ref_shape += overflow - underflow
```

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,70 @@
...
+        new_shape = np.rint(ref_shape * factor)
+        affine = rescale_affine(affine, ref_shape, ref_zooms / factor, new_shape)
+        ref_shape = new_shape
+
+    # Generate new reference
+    hdr.set_sform(affine, scode)
+    hdr.set_qform(affine, qcode)
+
+    return in_reference.__class__(
+        np.zeros(ref_shape.astype(int), dtype=hdr.get_data_dtype()),
```
effigies: You don't need a dtype for a sampling reference. In fact, you don't need to allocate the whole array, just something that has a shape and can be made an array on demand.

```suggestion
        nb.fileslice.strided_scalar(ref_shape.astype(int)),
```

effigies (commented):
The function looks good. Minor comments. I will have a detailed look at the test when I can, but it won't be today.

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,70 @@
 """Image processing tools."""
 
 
+def deoblique_and_zooms(in_reference, oblique, factor=4, padding=1):
```
mgxd: may as well type it since you made that nice docstring
```suggestion
import nibabel as nb

def deoblique_and_zooms(
    in_reference: nb.spatialimages.SpatialImage, 
    oblique: nb.spatialimages.SpatialImage, 
    factor: int = 4, 
    padding: int = 1
):
```

sdcflows/utils/tools.py
```diff
@@ -23,6 +23,70 @@
...
+        np.clip(extent_ijk.max(0) + padding + 1 - ref_shape, 0, None)
+    ).astype(int)
+    if np.any(underflow < 0) or np.any(overflow > 0):
+        # Add under/overflow voxels
+        ref_shape += np.abs(underflow) + overflow
+        # Consistently update origin
+        affine[:-1, -1] = apply_affine(affine, underflow)
+
+    # Make grid denser
+    if abs(1.0 - factor) > 1e-4:
```
mgxd: Why this magic number? Should this be a configurable threshold?

sdcflows/utils/tools.py
effigies: Note that `in_reference.header.get_qform()` is used, so `in_reference` needs to be `nb.Nifti1Image`, in case we ever want to type check this.

sdcflows/utils/tools.py
oesteban: Could be, but it's not that relevant. This is setting a tolerance for the factor to be considered 1. I could take it away or use epsilon from numpy.

oesteban: Okay, I added a function argument for this. You're right the best practice would be to make it configurable.

sdcflows/utils/tests/test_tools.py
oesteban: Added some parametric tests.

PR #358, opened 2023-05-11
 ENH: allow non-T1w anatomical estimators

eilidhmacnicol:
Anatomical reference images are less likely to be T1w in rodents. This would allow nipreps/fmriprep-rodents#55 the flexibility to provide a list of possible contrasts for anatomical images (in this case: T1w, T2w, UNIT1) to be used as an anatomical reference image without breaking compatibility for other use cases.

### DIFF
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -74,6 +74,9 @@ def find_estimators(
     bids_filters
         Optional dictionary of key/values to filter the entities on.
         This allows lower level file inclusion/exclusion.
+    anat_suffix : :obj:`str` or :obj:`list`
+        String or list of strings to filter anatomical images for fieldmap-less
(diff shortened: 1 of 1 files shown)


mgxd (dismissed):
Looks good to me

sdcflows/utils/wrangler.py
```diff
@@ -42,6 +42,7 @@ def find_estimators(
     force_fmapless: bool = False,
     logger: Optional[logging.Logger] = None,
     bids_filters: Optional[dict] = None,
+    anat_suffix: Union[str, list] = 'T1w',
```
effigies: Nit-pick:

```suggestion
    anat_suffix: Union[str, List[str]] = 'T1w',
```

effigies (approved):

PR #359, opened 2023-05-17
FIX: do not omit session bids_filter when listing b0 ids

bpinsard:
When running fMRIPrep on a single session (using bids filters), the wrangler will try to create fieldmap estimators from other session, and crash if data is not present (eg. datalad symlinks). 
This fix should prevent such crash.

### DIFF
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -317,9 +317,10 @@ def find_estimators(
             b0_entities = base_entities.copy()
             b0_entities["B0FieldIdentifier"] = b0_id
 
-            bare_ids = layout.get(**base_entities, B0FieldIdentifier=b0_id)
+            bare_ids = layout.get(**base_entities, session=sessions, B0FieldIdentifier=b0_id)
             listed_ids = layout.get(
                 **base_entities,
+                session=sessions,
                 B0FieldIdentifier=f'"{b0_id}"',  # Double quotes to match JSON, not Python repr
                 regex_search=True,
             )


bpinsard:
The error came from code I wrote, not realizing that session was popped from the filters.

Is is not clear if session should be also given to  https://github.com/nipreps/sdcflows/blob/master/sdcflows/utils/wrangler.py#L306 
Though it is now filtered after with the lines I changed.

I cannot imagine a case where a fieldmap from a different session should be used.

effigies:
In large, multi-session studies like ABCD, there are some good and some bad fieldmaps. Some groups have QC'd the fieldmaps and selected the best available, which they then apply across sessions. I think this would be best handled by https://github.com/nipreps/fmriprep/issues/2968, allowing sdcflows to find all fieldmaps and then only using relevant ones.

Alternately, in SDCFlows we could find all `B0FieldSource`s in the session and prune `B0FieldIdentifier`s that are not in the list.

bpinsard:
I understand that in some particular cases a fieldmap from a different session (with different subject-position, shimming,... ) is better than no fieldmap at all.
However, it's a bit disturbing that when specifically setting the bids_filters, these are not taken into account. 
I wonder if in the case of ABCD they specify the session in the BIDS filters while trying to get fieldmap in other sessions. I think that if not specified, it should look in all sessions and fit their case.

(I wonder if that is still a problem if it could be solved by reversing the search: list the B0FieldSource used in that session (or other bids filters) and then look for the files used to create that Fieldmap in the whole subject folder.) Oups, realized that this was what you suggested if I understood correctly.

effigies:
This is now handled in fMRIPrep. We may want to provide a filter mechanism here in sdcflows, to avoid duplicating code in other downstream users, but I don't think hiding fieldmaps from the wrangler is the right way to go about it.

bpinsard:
Unfortunately, this issue remains and had escaped my previous tests.
Even if now fmriprep do no run fieldmap estimation on fmaps not listed in b0 intents (https://github.com/nipreps/fmriprep/pull/3025) , the wrangler still lists/instantiates all fieldmap objects and crash when finding broken symlinks (datalad) on line https://github.com/nipreps/sdcflows/blob/master/sdcflows/fieldmaps.py#L202 . As we run jobs in parallel on temporary clones on the HPC, we only pull the strictly required data from the session to process, and thus it crashes.

An alternative would be to relax 
https://github.com/nipreps/sdcflows/blob/master/sdcflows/fieldmaps.py#L202 
and let broken symlink pass if only metadata are required to instantiate the fieldmap object ( which I am not sure about). 

In the reason mentioned no to apply session filters, the case where one wants to use fieldmaps from another session would still work by default if they provide no bids_filters for fmap or provide bids_filters that do not specify session, or include the other session, right?

PR #367, opened 2023-06-08
FIX: Ensure metadata is not present in entities

mgxd:
`BIDSImageFile.entities` populates both filename and metadata entities, which can result in an overzealous SQL join resulting in:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) at most 64 tables in a join
```

This change ensures metadata is not present when filtering.


effigies (approved):
Looks right.

PR #371, opened 2023-07-06
MAINT: Make call to scipy.stats.mode compatible with scipy 1.11.0

HippocampusGirl:
Starting with scipy 1.11.0, scipy.stats.mode returns a scalar value for the mode instead of a one-element array. This change was foreshadowed in scipy version 1.9.0 with the addition of the `keepdims` parameter, with the eventual goal of setting it to false by default for consistency with other scipy.stats functions.
This change is causing the _demean function to break, because it currently relies on indexing the resulting array.
```
NodeExecutionError: Exception raised while executing Node demean.

Traceback:
          File "/usr/local/miniconda/lib/python3.11/site-packages/nipype/interfaces/base/core.py", line 397, in run
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/usr/local/miniconda/lib/python3.11/site-packages/nipype/interfaces/utility/wrappers.py", line 142, in _run_interface
            out = function_handle(**args)
                  ^^^^^^^^^^^^^^^^^^^^^^^
          File "<string>", line 26, in _demean
        IndexError: invalid index to scalar variable.
```
See https://github.com/scipy/scipy/issues/16418 for the initial proposal and https://github.com/scipy/scipy/pull/17561 for the final change that is now breaking the _demean function.

This commit changes the indexing to be compatible with scipy 1.11.0 and other future versions. I have also tested this change with older scipy versions up to 0.17.1, which was released in 2016.

### DIFF
--- a/sdcflows/workflows/gre.py
+++ b/sdcflows/workflows/gre.py
@@ -212,7 +212,7 @@ def _demean(in_file, in_mask=None, usemode=True):
 
     if usemode:
         from scipy.stats import mode
-        data[msk] -= mode(data[msk], axis=None)[0][0]
+        data[msk] -= mode(data[msk], axis=None).mode.item()
     else:
         data[msk] -= np.median(data[msk], axis=None)
 


HippocampusGirl:
The CI error seems to be a FreeSurfer license issue, so not related to the proposed change.
```
ERROR: FreeSurfer license file /opt/freesurfer/license.txt not found.
```

effigies:
Pushed to a local branch to ensure CI has access to any secrets needed. https://app.circleci.com/pipelines/github/nipreps/sdcflows/1543/workflows/8fda487f-2ebc-46d4-82e5-eb99b8ec638f / #372.

PR #375, opened 2023-07-07
MNT: Add pre-commit config

effigies:
The motivator is to avoid `import pdb; pdb.set_trace()` and `breakpoint()`s from getting into code. I picked a selection of the default hooks.

I don't want to interfere with #346, so I left black and isort commented, but we've adopted these in fmriprep, and I think it's reasonable to do so here.


mgxd (approved):
just noting to run `pre-commit install` in dev environments :)

PR #381, opened 2023-07-12
ENH: Output target2fmap_xfm from coeff2epi_wf

effigies:
Stealing a bit from #346 because I need it for fit-apply.

### DIFF
--- a/sdcflows/workflows/apply/registration.py
+++ b/sdcflows/workflows/apply/registration.py
@@ -87,6 +87,10 @@ def init_coeff2epi_wf(
     target_ref
         the target reference EPI resampled into the fieldmap reference for
         quality control purposes.
+    target2fmap_xfm
+        An ITK-style transform produced with ANTs that can be used as transform in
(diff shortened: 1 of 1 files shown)


sdcflows/workflows/apply/registration.py
```diff
@@ -133,14 +138,17 @@ def init_coeff2epi_wf(
             ("target_mask", f"moving_image_mask{mask_trait_s}"),
             ("fmap_mask", f"fixed_image_mask{mask_trait_s}"),
         ]),
-        (coregister, outputnode, [("warped_image", "target_ref")]),
+        (coregister, outputnode, [
+            ("warped_image", "target_ref"),
```
oesteban, line 142: This would require changing line 123 to `True`, unless you only want the reference when debugging.

oesteban (approved):
LGTM, I just left a cautionary note.

sdcflows/workflows/apply/registration.py
effigies: That's fine. That's not new, all I want for now is the transform.

PR #385, opened 2023-08-04
ENH: Allow unwarp to apply only to references

effigies:
This PR adds a `ref_only` mode to `init_unwarp_wf` that allows the same workflow to be used only on reference volumes.

This also changes the output reference volume to be derived from the input reference volume, rather than the average of the corrected time series.

### DIFF
--- a/sdcflows/workflows/apply/correction.py
+++ b/sdcflows/workflows/apply/correction.py
@@ -26,7 +26,7 @@ from nipype.interfaces import utility as niu
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 
 
-def init_unwarp_wf(*, free_mem=None, omp_nthreads=1, debug=False, name="unwarp_wf"):
+def init_unwarp_wf(*, ref_only=False, free_mem=None, omp_nthreads=1, debug=False, name="unwarp_wf"):
     r"""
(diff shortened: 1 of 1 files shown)


effigies:
This is intended to reduce the footprint of the EPI fit workflow. The corrected reference image will be used for T1w coregistration.

oesteban (dismissed):
This changes the idea that a new reference is generated with the unwarped data, but that is probably is good overall, as the programmer will be required to generate the new reference explicitly with the undistorted data, if they don't want to directly use the unwarped reference.

Looks great to me :)

oesteban:
I came back to check one thing - the `distorted_ref` should be a realigned file if it has more than 1 timepoints (I am currently working with sbrefs having 4 volumes)

effigies:
Sorry, you mean distorted_ref should be averaged, in case it has multiple time points? I thought by the time we got to this, sbrefs had already been merged to a single volume.

oesteban:
> Sorry, you mean distorted_ref should be averaged, in case it has multiple time points? I thought by the time we got to this, sbrefs had already been merged to a single volume.

If that is the expectation, we should make it very clear that it has been averaged. However, resample would handle that just fine, allowing less data munging before correcting.

effigies:
Closing in favor of #346. We will probably want to make averaging and brain extraction optional in a future PR, to avoid duplicating effort when only a timeseries is needed.

PR #388, opened 2023-08-17
FIX: Clean up ``scipy.signal.cubic`` deprecation

### LINKED ISSUE #370: ``scipy.signal.cubic`` is now deprecated
### What happened?

https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.cubic.html#scipy.signal.cubic

### What command did you use?

```shell
n/a
```

### What version of the software are you running?

n/a

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

No

_No response_

_No response_

oesteban:
Resolves: #370.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -56,8 +56,8 @@ import attr
 import numpy as np
 from warnings import warn
 from scipy import ndimage as ndi
-from scipy.signal import cubic
-from scipy.sparse import vstack as sparse_vstack, kron, lil_array
+from scipy.interpolate import BSpline
+from scipy.sparse import vstack as sparse_vstack, kron
 
 import nibabel as nb
 import nitransforms as nt
@@ -325,7 +325,12 @@ class B0FieldTransform:
         for level in coeffs:
             wmat = grid_bspline_weights(target_reference, level)
             weights.append(wmat)
-            coeffs_data.append(level.get_fdata(dtype="float32").reshape(-1))
+
+            # Coefficients must be zero-padded because Scipy's BSpline pads the knots.
+            coeffs_data.append(np.pad(
+                level.get_fdata(dtype="float32"),
+                ((1, 1), (1, 1), (1, 1)),
+            ).reshape(-1))
 
         # Reconstruct the fieldmap (in Hz) from coefficients
         fmap = np.zeros(projected_reference.shape[:3], dtype="float32")
@@ -732,18 +737,16 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
         coords[axis] = np.arange(sample_shape[axis], dtype=dtype)
 
         # Calculate the index component of samples w.r.t. B-Spline knots along current axis
+        # Size of locations is L
         locs = nb.affines.apply_affine(target_to_grid, coords.T)[:, axis]
-        knots = np.arange(knots_shape[axis], dtype=dtype)
 
-        distance = np.abs(locs[np.newaxis, ...] - knots[..., np.newaxis])
-        within_support = distance < 2.0
-        d_vals, d_idxs = np.unique(distance[within_support], return_inverse=True)
-        bs_w = cubic(d_vals)
+        # Size of knots is K + 6 so that all locations are fully covered by basis
+        knots = np.arange(-3, knots_shape[axis] + 3, dtype=dtype)
 
-        colloc_ax = lil_array((knots_shape[axis], sample_shape[axis]), dtype=dtype)
-        colloc_ax[within_support] = bs_w[d_idxs]
-
-        wd.append(colloc_ax)
+        # Our original design matrices had size (K, L)
+        # However, BSpline.design_matrix() generates a size of (L, K + 2),
+        # hence the trasposition (and zero-padding of 1 at every face when using these)
+        wd.append(BSpline.design_matrix(locs, knots, 3).T)
 
     # Calculate the tensor product of the three design matrices
     return kron(kron(wd[0], wd[1]), wd[2]).astype(dtype)
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -745,7 +745,7 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
 
         # Our original design matrices had size (K, L)
         # However, BSpline.design_matrix() generates a size of (L, K + 2),
-        # hence the trasposition (and zero-padding of 1 at every face when using these)
+        # hence the transposition (and zero-padding of 1 at every face when using these)
         wd.append(BSpline.design_matrix(locs, knots, 3).T)
 
     # Calculate the tensor product of the three design matrices
--- a/sdcflows/tests/test_transform.py
+++ b/sdcflows/tests/test_transform.py
@@ -348,7 +348,16 @@ def test_grid_bspline_weights():
         nb.Nifti1Image(np.zeros(target_shape), target_aff),
         nb.Nifti1Image(np.zeros(ctrl_shape), ctrl_aff),
     ).tocsr()
-    assert weights.shape == (64, 1000)
+    assert weights.shape == (np.prod(np.array(ctrl_shape) + 2), np.prod(target_shape))
+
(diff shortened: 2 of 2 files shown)


sdcflows/transform.py
```diff
@@ -732,18 +737,16 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
...
 
-        colloc_ax = lil_array((knots_shape[axis], sample_shape[axis]), dtype=dtype)
-        colloc_ax[within_support] = bs_w[d_idxs]
+        # Size of knots is K + 6 so that all locations are fully covered by basis
+        knots = np.arange(-3, knots_shape[axis] + 3, dtype=dtype)
 
-        wd.append(colloc_ax)
+        # Our original design matrices had size (K, L)
+        # However, BSpline.design_matrix() generates a size of (L, K + 2),
+        # hence the trasposition (and zero-padding of 1 at every face when using these)
```
```suggestion
        # hence the transposition (and zero-padding of 1 at every face when using these)
```

sdcflows/transform.py
```diff
@@ -732,18 +737,16 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
         coords[axis] = np.arange(sample_shape[axis], dtype=dtype)
 
         # Calculate the index component of samples w.r.t. B-Spline knots along current axis
+        # Size of locations is L
         locs = nb.affines.apply_affine(target_to_grid, coords.T)[:, axis]
-        knots = np.arange(knots_shape[axis], dtype=dtype)
-
-        distance = np.abs(locs[np.newaxis, ...] - knots[..., np.newaxis])
-        within_support = distance < 2.0
```
effigies, line 739: IIRC this heuristic is what really saved us the time and memory.

sdcflows/transform.py
```diff
@@ -732,18 +737,16 @@ def grid_bspline_weights(target_nii, ctrl_nii, dtype="float32"):
...
-        colloc_ax = lil_array((knots_shape[axis], sample_shape[axis]), dtype=dtype)
-        colloc_ax[within_support] = bs_w[d_idxs]
+        # Size of knots is K + 6 so that all locations are fully covered by basis
+        knots = np.arange(-3, knots_shape[axis] + 3, dtype=dtype)
 
-        wd.append(colloc_ax)
+        # Our original design matrices had size (K, L)
+        # However, BSpline.design_matrix() generates a size of (L, K + 2),
+        # hence the trasposition (and zero-padding of 1 at every face when using these)
+        wd.append(BSpline.design_matrix(locs, knots, 3).T)
```
effigies, line 749: Does it make sense to drop the extra knots, rather than pad the coefficients?

```suggestion
        wd.append(BSpline.design_matrix(locs, knots, 3)[:, 1:-1].T)
```

sdcflows/transform.py
oesteban: Makes sense - should we try to investigate how `cubic()` has been internally implemented?

sdcflows/transform.py
oesteban: Yeah, I thought about this, and it may be positive to add a frame of zeros in the coefficients map. That way we may avoid weird border effects when coefficients at the edge are very large.

sdcflows/transform.py
effigies: Seems to be via fitpack: https://github.com/scipy/scipy/blob/v1.11.1/scipy/interpolate/_bspl.pyx#L466

I think more importantly, the locations evaluated seems to be x * k:

https://github.com/scipy/scipy/blob/cfe80116aaa145061246b8aec0e98248fbefb835/scipy/interpolate/_bsplines.py#L445-L453

I don't fully understand why this is a CSR matrix, it looks dense to me, but I might be missing something.

effigies:
I've read and re-read the BSpline docs, and I cannot for the life of me figure out how we're supposed to get the derivatives to generate the Jacobian. Do you know how that's supposed to work?

oesteban:
I guess https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.derivative.html

Something like `BSpline.derivative().design_matrix(...)` ?

oesteban:
The coefficients are the same. It's just the weights that change. Pretty convenient :)

effigies:
`design_matrix()` is a classmethod. If we find a way to write `generate_design_matrix(BSpline())`, then we could do `generate_design_matrix(BSpline().derivative())`.

oesteban:
I see. We can initialize a `BSpline` because `t`, `c` and `k` are known when we want to calculate the jacobian. However, we will need to see how the design matrix is generated for an object because we need to calculate the tensor product.

oesteban:
Now I'm hoping my Cython implementation is still in my local git repository...

effigies:
Well, they do have an equivalence in https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.design_matrix.html#scipy.interpolate.BSpline.design_matrix:

```python
from scipy.interpolate import make_interp_spline, BSpline
import numpy as np

k = 2
t = [-1, 0, 1, 2, 3, 4, 5, 6]
x = [1, 2, 3, 4]
c = np.eye(len(t) - k - 1)

design_matrix = BSpline.design_matrix(x, t, k)
design_matrix_gh = BSpline(t, c, k)(x)
np.allclose(design_matrix.toarray(), design_matrix_gh, atol=1e-14)
```

So we could try:

```Python
orig = BSpline(knots, np.eye(len(knots) - 3 - 1), 3)
wd0.append(scipy.sparse.csr_matrix(orig(locs).T))

if jacobian:
    deriv = orig.derivative()
    wd1.append(scipy.sparse.csr_matrix(deriv(locs).T))
```

oesteban:
> Well, they do have an equivalence in https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.BSpline.design_matrix.html#scipy.interpolate.BSpline.design_matrix:
> 
> ```python
> from scipy.interpolate import make_interp_spline, BSpline
> import numpy as np
> 
> k = 2
> t = [-1, 0, 1, 2, 3, 4, 5, 6]
> x = [1, 2, 3, 4]
> c = np.eye(len(t) - k - 1)
> 
> design_matrix = BSpline.design_matrix(x, t, k)
> design_matrix_gh = BSpline(t, c, k)(x)
> np.allclose(design_matrix.toarray(), design_matrix_gh, atol=1e-14)
> ```
> 
> So we could try:
> 
> ```python
> orig = BSpline(knots, np.eye(len(knots) - 3 - 1), 3)
> wd0.append(scipy.sparse.csr_matrix(orig(locs).T))
> 
> if jacobian:
>     deriv = orig.derivative()
>     wd1.append(scipy.sparse.csr_matrix(deriv(locs).T))
> ```

I don't see why this would not work.

sdcflows/transform.py
oesteban: BTW - we probably can get rid of the padding problem by setting `extrapolate=True` when creating the design_matrix. I'm not 100% positive that is a great idea though.

sdcflows/transform.py
```diff
@@ -325,7 +325,12 @@ def fit(
         for level in coeffs:
             wmat = grid_bspline_weights(target_reference, level)
             weights.append(wmat)
-            coeffs_data.append(level.get_fdata(dtype="float32").reshape(-1))
+
+            # Coefficients must be zero-padded because Scipy's BSpline pads the knots.
+            coeffs_data.append(np.pad(
+                level.get_fdata(dtype="float32"),
+                ((1, 1), (1, 1), (1, 1)),
+            ).reshape(-1))
```
effigies, line 333: If we're not removing weights when fitting, the coefficients should match the size of the design matrix, so we should not need to pad them here, correct?

oesteban:
I think it's the opposite, if I understood your message correctly. If we
don't drop weights, we will need to pad here.

effigies:
Math is hard, and I should probably trust the tests. If you're comfortable that we're testing everything thoroughly, I'm okay merging. I'm working on a branch off of here to add Jacobians based on my above comment.

effigies:
Replaced by #393.

PR #391, opened 2023-08-18
ENH: Implement Jacobian weighting during unwarp

### LINKED ISSUE #382: Implement Jacobian attenuation in the target space
### What would you like to see added in this software?

SDCFlows should provide a method to generate a Jacobian weight image to multiply unwarped images by.

Like the fieldmap estimate, the Jacobian should be sampled into the target space, so that it is a simple post-resampling multiplication.

To reduce copies and compounding floating-point error (for example, if we resample as float64 and save as float32), ideally our resampler would also accept a Jacobian and multiply before writing to disk.

### Do you have any interest in helping implement the feature?

Yes

_No response_

effigies:
This turned out to be dead simple. The basic idea of a Jacobian is to map volumes in the target space to volumes in the source space. We have a particularly easy case, because there is a warp only in one dimension, so our Jacobian is trivially equal to $1 + dx/dP$ where $P$ is the phase-encoding axis.

It seems likely that we will need to adjust this when permitting other transformations. While affine transforms will scale all points uniformly (and can thus be ignored for unitless values like BOLD), warps to template will almost certainly produce additional changes worth accounting for, though the scale of those effects is probably small compared to SDC.

Closes #382.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -92,7 +92,15 @@ def _sdc_unwarp(
     # Map voxel coordinates applying the VSM
     # The VSM is just the displacements field given in index coordinates
     # voxcoords is the deformation field, i.e., the target position of each voxel
-    coordinates[pe_info[0], ...] += fmap_hz * pe_info[1]
+    vsm = fmap_hz * pe_info[1]
+    coordinates[pe_info[0], ...] += vsm
+
+    # The Jacobian determinant image is the amount of stretching in the PE direction.
+    # Using central differences accounts for the shift in neighboring voxels.
+    # The full Jacobian at each voxel would be a 3x3 matrix, but because there is
+    # only warping in one direction, we end up with a diagonal matrix with two 1s.
+    # The following is the other entry at each voxel, and hence the determinant.
+    jacobian = 1 + np.gradient(vsm, axis=pe_info[0])
 
     resampled = ndi.map_coordinates(
         data,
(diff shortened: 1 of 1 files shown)


effigies:
This definitely doesn't improve things:

image

effigies:
~~Hmm. I wonder if I should be scaling the jacobian by the voxel volume, not from 0-1...~~ (This was not even wrong.)

effigies:
Continuing to work on this. I have attempted to create a Jacobian determinant in a couple ways, none of which significantly affected my use case. I believe that the mask used to fit the spline field is the more significant problem for phasediff fieldmaps. I think we need to mask the output field.

effigies:
~~As discussed with @oesteban, the shortest path forward here is to pull the TOPUP-generated Jacobians to the application stage. Calculating our own Jacobians can wait.~~

Turns out trying to make use of TOPUPs Jacobians was more pain than finally wrapping my head around exactly what a Jacobian is in this context.

effigies:
#### `applytopup`-corrected

Screenshot from 2023-10-01 00-33-32

#### `sdcflows @ master`-corrected
Screenshot from 2023-10-01 00-33-58

#### With this PR
Screenshot from 2023-10-01 00-34-08

It is not 100% identical, and this may be a result of premature coercion to float32 of coordinates during resampling, or it could be our spline knots are very slightly differently defined.

oesteban:
Excellent work 💯

effigies:
@oesteban or @mgxd would you care to review? (Feel free to say you don't have time but I should/shouldn't go ahead and merge.)

mgxd (approved):
this is great and lgtm - let's get it into dev for further testing

PR #393, opened 2023-09-22
RF: Use scipy.interpolate.BSpline to construct spline basis

effigies:
This builds on and replaces #388. The additional padding broke our ability to reconstruct the field, as we were estimating more parameters but not actually saving them.

This PR maintains the current functionality and paves the way for implementing the Jacobian (#391) by using `BSpline()` objects directly instead of through `BSpline.design_matrix()`. I also restore the use of a `within_support` mask, which significantly decreases the runtime of `unwarp_wf.resample` nodes. There is little effect on `bs_filter` nodes, as we downsample to a low enough resolution to make the fit process very quick already.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -56,8 +56,8 @@ import attr
 import numpy as np
 from warnings import warn
 from scipy import ndimage as ndi
-from scipy.signal import cubic
-from scipy.sparse import vstack as sparse_vstack, kron, lil_array
+from scipy.interpolate import BSpline
+from scipy.sparse import vstack as sparse_vstack, kron
 
 import nibabel as nb
 import nitransforms as nt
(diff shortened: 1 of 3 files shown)


effigies:
@oesteban This should be a relatively quick review. Apart from the specific B-spline invocation change, the main change is that our design matrices are transposed from before. The effect is that we can stop transposing them after generation.

The tests verify that the `kron()` products are unchanged apart from the transposition.

sdcflows/transform.py
```diff
@@ -321,17 +320,15 @@ def fit(
             )
 
         # Generate tensor-product B-Spline weights
-        coeffs_data = []
-        for level in coeffs:
-            wmat = grid_bspline_weights(target_reference, level)
-            weights.append(wmat)
-            coeffs_data.append(level.get_fdata(dtype="float32").reshape(-1))
+        colmat = sparse_hstack(
+            [grid_bspline_weights(target_reference, level) for level in coeffs]
```
```suggestion
            [grid_bspline_weights(projected_reference, level) for level in coeffs]
```

effigies:
Read through with fresh eyes. I'm going to go ahead and merge.

PR #395, opened 2023-09-26
FIX: Mask B0 field estimate during application

effigies:
This addresses issues like nipreps/fmriprep#3013. For phasediff images, the field estimate outside the mask is not used to generate the B-spline representation, which means that the error outside the mask can be arbitrarily large. This results in pulling EPI signal to regions well outside the brain and smearing signal on the edge of the brain.

This PR allows the estimation mask to be passed to correction.

This requires an additional PR in fMRIPrep to be connected up properly. It's possible that the connection will need to be conditional on fieldmap type, as not all types use the mask for fitting the splines.

Big thanks to @feilong for helping point me in the right direction here.

Builds on #393.

### Before
image

### After
image


effigies:
Here's an alternative, where instead of masking during application, we mask the fieldmap before fitting the spline estimate:

image

This seems to produce less of a discontinuity, although it doesn't fix similar problems for PEPOLAR scans.

effigies:
After discussion with @oesteban, we've settled on the approach in #396 instead.

PR #396, opened 2023-09-27
FIX: Mask fieldmap before fitting spline field

effigies:
Previously, we fit a spline field to the within-mask portion of a fieldmap, which could lead to large peaks outside the mask. Applying the mask to the reconstructed field can produce discontinuities in resampling. Instead of attempting to attenuate the peaks outside the mask in a smooth way, we set the fit values outside the mask to zero, and let the spline fit find a smooth field.

Replaces #395.
Addresses nipreps/fmriprep#3013.

## Before

Screenshot from 2023-09-26 07-48-16

## After

Screenshot from 2023-09-26 10-39-02

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -195,6 +195,7 @@ class BSplineApprox(SimpleInterface):
             center = np.mean(data[mask])
 
         data -= center
+        data[~mask] = 0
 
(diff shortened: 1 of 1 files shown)


effigies:
As discussed, going ahead with this option. This will be in the next unstable release of fmriprep.

PR #397, opened 2023-10-04
ENH: Implement Jacobian weighting during unwarp [backport gh-391]

effigies:
Slightly different shaping than in the latest `master`, but I believe this is the equivalent. It would be good to get a test. @themeo would you be able to test? If you're using `fmriprep-docker`, it's pretty easy:

```console
$ git clone https://github.com/effigies/sdcflows.git
$ git -C sdcflows checkout bp/2.5.x/391
$ pipx run "fmriprep-docker==23.1.4" [usual arguments] --patch sdcflows=$PWD/sdcflows/sdcflows
```

This will mount the sdcflows directory into the correct location to be picked up by fMRIPrep.

### DIFF
--- a/sdcflows/transform.py
+++ b/sdcflows/transform.py
@@ -197,7 +197,10 @@ class B0FieldTransform:
         # fmap * ro_time is the voxel-shift map (VSM)
         # The VSM is just the displacements field given in index coordinates
         # voxcoords is the deformation field, i.e., the target position of each voxel
-        voxcoords[pe_axis, ...] += fmap.reshape(-1) * ro_time
+        vsm = fmap * ro_time
+        voxcoords[pe_axis, ...] += vsm.reshape(-1)
+
+        jacobian = 1 + np.gradient(vsm, axis=pe_axis)
 
         # Prepare data
         data = np.squeeze(np.asanyarray(spatialimage.dataobj))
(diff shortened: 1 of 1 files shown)


themeo:
I'd be happy to give it a spin, but our cluster doesn't have docker (we use singularity). Is it possible to put the patched version together using singularity?

effigies:
Yes, it's just a little bit less automated. You'll need to add the following flag to your binds, before the image:

```
-B $PWD/sdcflows/sdcflows:/opt/conda/envs/fmriprep/lib/python3.10/site-packages/sdcflows
```

You can replace `$PWD/sdcflows` with the full path to wherever you clone the repository. Note that on the left side of the colon `sdcflows` should appear twice, like `sdcflows/sdcflows:...`

themeo:
This fix led to a dramatic improvement in how fmriprep handles SDC. 

original raw bold:
image

topup/applytopup on the raw image:
image

fmriprep with the fix:
image

This is not an entirely clean comparison, as the fmriprep-processed image has also undergone motion correction etc. 

Congrats on cracking this!

amirhusseinab:
Thanks for the fix! The boldref is SDCed now, but the T1w and MNI space outputs are still extra-distorted. Is it because the pipelines are not totally merged? PE-dir is RL, SDC with PEPOLAR.
Raw bold: 
image

fmriprep 23.1.4. "vannila" boldref:
image

fmriprep 23.1.4. SDC-fixed boldref:
image

fmriprep 23.1.4 vannila and SDC-fixed T1w space boldref:
image

effigies:
This is a good point, this will currently only affect files resampled into boldref space, since the resampling to other spaces is handled differently. I'm currently working on a tool that will resample directly to any volumetric space, but that will only be available with fmriprep 23.2.

effigies:
This will need to wait for a later fMRIPrep.

PR #398, opened 2023-10-04
FIX: Mask fieldmap before fitting spline field [backport gh-396]

effigies:
Straightforward backport. Would still be good to get a test. To anybody willing:

```
$ git clone https://github.com/effigies/sdcflows.git
$ git -C sdcflows checkout bp/2.5.x/396
$ pipx run "fmriprep-docker==23.1.4" [usual arguments] --patch sdcflows=$PWD/sdcflows/sdcflows
```

Backport of #396.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -195,6 +195,7 @@ class BSplineApprox(SimpleInterface):
             center = np.mean(data[mask])
 
         data -= center
+        data[~mask] = 0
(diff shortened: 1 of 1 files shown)


effigies:
@oesteban WDYT about this backport? Are you comfortable doing this in a bug-fix release? I'm considering dropping it in and making a final bug-fix release of 23.1.x so the fix can be available for people before 23.2 is finalized

PR #408, opened 2023-11-28
ENH: Standalone CLI now estimates fieldmaps

oesteban:
The previous behavior (just "finding" estimators) is maintained with the argument ``--dry-run``.


oesteban:
Well, at df93c7b, this is surprisingly executing the estimation of GRE fieldmaps. EPI fieldmaps failed in the robust average, so I'll need to check what went wrong.

.github/workflows/docs-build-pr.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     - name: Set up Python 3
       uses: actions/setup-python@v4
       with:
-        python-version: 3
+        python-version: "3.10"
```
effigies: What's the need for an older version? 3 was specifically intended to avoid pinning. The default Python action marches along at a reasonably slow pace, so we shouldn't need to worry about 3.12 becoming default before the ecosystem is ready, but by now it should be safe to build on 3.11.

docs/_static/config-example.toml
```diff
@@ -0,0 +1,53 @@
...
+output_dir = "/home/oesteban/tmp/sdcflows-hcph/out"
+participant_label = [ "001",]
+pdb = false
+run_uuid = "20231129-112254_e9d6d38e-35ff-463e-9f7e-1ff15ecd54ac"
+templateflow_home = "/data/templateflow"
+work_dir = "/home/oesteban/tmp/sdcflows-hcph/work"
+write_graph = false
+
+[workflow]
+analysis_level = [ "participant",]
```
effigies, line 33: This is expected to be a list?

sdcflows/cli/main.py
```diff
@@ -0,0 +1,182 @@
...
+        from sdcflows.utils.telemetry import setup_migas
+
+        setup_migas()
+
+    # CRITICAL Call build_workflow(config_file, retval) in a subprocess.
+    # Because Python on Linux does not ever free virtual memory (VM), running the
+    # workflow construction jailed within a process preempts excessive VM buildup.
+    from multiprocessing import Manager, Process
+
+    with Manager() as mgr:
```
effigies, line 122: Is it needed to separate this, if you've already initialized the process pool?

sdcflows/cli/main.py
```diff
@@ -0,0 +1,182 @@
...
+                    pe_dir = fl.metadata.get("PhaseEncodingDirection")
+                    print(f"\t\t\t{pe_dir}\t{fl_relpath}")
+        sys.exit(exitcode)
+
+    # Initialize process pool if multiprocessing
+    _pool = None
+    if config.nipype.plugin in ("MultiProc", "LegacyMultiProc"):
+        from contextlib import suppress
+        import multiprocessing as mp
+        import multiprocessing.forkserver
```
effigies: Unused?

.github/workflows/docs-build-pr.yml
oesteban: Not sure why I did this -- rolling back

```suggestion
        python-version: 3
```

.github/workflows/docs-build-update.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     - name: Set up Python 3
       uses: actions/setup-python@v4
       with:
-        python-version: 3
+        python-version: "3.10"
```
oesteban:
```suggestion
        python-version: 3
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+    )
+
+    # main options
+    g_bids = parser.add_argument_group("Options related to BIDS")
+    g_bids.add_argument(
+        "--participant-label",
+        "--participant_label",
+        "--participant-labels",
+        "--participant_labels",
+        dest="participant_label",
```
effigies: I would be inclined not to have synonyms when they aren't required for backwards compatibility.

```suggestion
        "--participant-label",
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        action=ParticipantLabelAction,
+        nargs="+",
+        help="A space delimited list of participant identifiers or a single "
+        "identifier (the sub- prefix can be removed).",
+    )
+    g_bids.add_argument(
+        "--session",
```
effigies: For consistency

```suggestion
        "--session-label",
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        help="Wipe out previously existing BIDS indexing caches, forcing re-indexing.",
+    )
+
+    # General performance
+    g_perfm = parser.add_argument_group("Options to handle performance")
+    g_perfm.add_argument(
+        "--nprocs",
+        "--n_procs",
+        "--n_cpus",
+        "-n-cpus",
```
effigies: Again, I would pick one and go with it. I don't care which.

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+8-CPU-threaded processes will be running at a given time). \
+Under this mode of operation, ``--nprocs`` sets the maximum number of processors \
+that can be assigned work within an *SDCFlows* job, which includes all the processors \
+used by currently running single- and multi-threaded processes. \
+If ``None``, the number of CPUs available will be automatically assigned (which may \
+not be what you want in, e.g., shared systems like a HPC cluster.""",
+    )
+    g_perfm.add_argument(
+        "--omp-nthreads",
+        "--ants-nthreads",
```
effigies:
```suggestion
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        type=PositiveInt,
+        help="""\
+Maximum number of threads that multi-threaded processes executed by *SDCFlows* \
+(e.g., ANTs' registration) can use. \
+If ``None``, the number of CPUs available will be automatically assigned (which may \
+not be what you want in, e.g., shared systems like a HPC cluster.""",
+    )
+    g_perfm.add_argument(
+        "--mem",
+        "--mem_gb",
```
effigies:
```suggestion
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        dest="memory_gb",
+        action="store",
+        type=_to_gb,
+        help="Upper bound memory limit for SDCFlows processes.",
+    )
+    g_perfm.add_argument(
+        "--debug",
+        action="store_true",
+        default=False,
+        help="Initiate in debug mode",
```
effigies:
```suggestion
        help="Enable changes to processing to aid in debugging",
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        "-n",
+        "--dry-run",
+        action="store_true",
+        default=False,
+        help="only find estimable fieldmaps (that is, estimation is not triggered)",
+    )
+    g_outputs.add_argument(
+        "--fmapless",
+        action="store_true",
+        default=False,
```
effigies: IMO it should be considered a first-class option. It's legitimate to disable it, but I think we should default to True.

```suggestion
        "--no-fmapless",
        action="store_false",
        dest="fieldmapless",
        default=False,
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        default=None,
+        type=Path,
+        help="Nipype plugin configuration file.",
+    )
+    g_outputs.add_argument(
+        "--notrack",
+        action="store_true",
+        help="Opt-out of sending tracking information of this run to the NiPreps developers. This"
+        " information helps to improve SDCFlows and provides an indicator of real world usage "
+        " crucial for obtaining funding.",
```
effigies: Spacing is off. Adding a semantic line break.

```suggestion
        help="Opt-out of sending tracking information of this run to the NiPreps developers. "
        "This information helps to improve SDCFlows and provides an indicator of "
        "real world usage for obtaining funding.",
```

sdcflows/cli/parser.py
```diff
@@ -0,0 +1,356 @@
...
+        "participant level analysis.",
+    )
+    parser.add_argument(
+        "analysis_level",
+        action="store",
+        nargs="+",
+        help="Level of the analysis that will be performed. "
+        "Multiple participant level analyses can be run independently "
+        "(in parallel) using the same output_dir.",
+        choices=["participant", "group"],
```
effigies, line 120: What is a group-level analysis?

docs/_static/config-example.toml
oesteban: yup (for laziness). It *could* be `[ "participant", "group" ]`, however group currently doesn't have an implementation (perhaps generating a group report?)

sdcflows/cli/main.py
oesteban: We could send it to the pool, that'd be fine. However, it's working this way in MRIQC.

sdcflows/cli/main.py
oesteban:
```suggestion
```

sdcflows/cli/parser.py
oesteban: Right now, it's not a thing.

sdcflows/cli/parser.py
oesteban:
```suggestion
```

sdcflows/config.py
```diff
@@ -0,0 +1,669 @@
...
+    """Disable the sharing of usage information with developers."""
+    output_dir = None
+    """Folder where derivatives will be stored."""
+    participant_label = None
+    """List of participant identifiers that are to be preprocessed."""
+    pdb = False
+    """Drop into PDB when exceptions are encountered."""
+    run_uuid = "{}_{}".format(strftime("%Y%m%d-%H%M%S"), uuid4())
+    """Unique identifier of this particular run."""
+    session = None
```
oesteban:
```suggestion
    session_label = None
```

sdcflows/cli/main.py
effigies: My point was the pool is initialized, so isn't it safe to do this in the main process at this point? These managers tend to make debugging trickier, so if we can avoid it, I'd rather.

effigies (approved):
Go ahead and merge this when you're ready. It would be nice if we could get away from the `Manager()` indirection for building workflows eventually, but I don't want to hold this up on that.

sdcflows/cli/main.py
oesteban: Oh, yes, in theory we could use the main process without problems, indeed.

oesteban: Okay, I was removing this context to build the workflow in the main thread, and it could be interesting to keep it this way, as the workflow is created while OMP_NUM_THREADS is set to 1. This saves great numpy buildup. 

We could still move it to the main process, but then it would make sense to make sure numpy is imported with OMP_NUM_THREADS = 1 (and that is then overridden in  subprocesses).

Issue #405, opened 2023-11-15
Make suffixes and datatypes configurable, if possible


tsalo:
### What would you like to see added in this software?

I am trying to replace an internal, outdated copy of SDCFlows in ASLPrep with the current version. However, since `asl` and `m0scan` are not supported suffixes for PEPOLAR or fieldmap-less distortion correction, I cannot use SDCFlows as a dependency without making changes to the codebase, getting those changes merged in, and getting a new release minted.

It seems like this restriction could be alleviated if suffixes and datatypes could be configured outside of the functions and classes.

### Do you have any interest in helping implement the feature?

Yes

### Additional information / screenshots

For example, there's a constant in `sdcflows.fieldmaps` called `MODALITIES` that I could override with a context manager when I use `FieldmapFile` in ASLPrep:

https://github.com/nipreps/sdcflows/blob/8ca72803b10f2114482029d06680f4a48215cc6a/sdcflows/fieldmaps.py#L54-L68

However, `FieldmapFile` then has suffixes hardcoded throughout, so, even if I modified `MODALITIES` it still wouldn't recognize the suffixes I need. For example:

https://github.com/nipreps/sdcflows/blob/8ca72803b10f2114482029d06680f4a48215cc6a/sdcflows/fieldmaps.py#L243-L248

effigies:
Sounds completely reasonable. I think it will be worth sitting down and thinking through the API at some point, but it does not sound like this will significantly modify our API right now, so full steam ahead.

tsalo:
Once I get ASLPrep working with my branch of SDCFlows, I'll be able to open a PR.

oesteban:
This would be worth bringing up in a TechMon meeting.

@tsalo - the next one will be focused on multi-echo fMRIPrep, but we may have some time for this (at least comment it out so it reaches beyond those who read this issue).

PR #411, opened 2023-12-01
ENH: Add support for ASL data

tsalo:
Closes none, but is related to #405. I'm happy to take a crack at adding tests, but it would be good to know what the preferred approach is. The easiest thing might be for me to just copy existing test files and rename them. I don't think distortion correction of EPI ASL will be any different from SDC of other EPI sequences, but I could add actual ASL and M0 data if that would be better. Unfortunately, I don't have access to data from a number of the cases listed below.

There are a few cases that I think should be covered. Do they make sense?

1. ASL scan with AP and PA M0 scans located in `fmap`, with separate M0 scan in `perf`.
2. AP ASL scan with PA M0 scan in `fmap`, with separate M0 scan in `perf`.
3. AP and PA ASL scans (is that supported yet?), with separate M0 scan in `perf`.
4. ASL scan with phase-difference field maps, with separate M0 scan in `perf`.
5. ASL scan without any field map.

Changes proposed:

- Add `asl` and `m0scan` to PEPOLAR and fieldmap-less suffix lists.
    - I essentially added "asl" wherever I saw "dwi".


sdcflows/utils/wrangler.py
```diff
@@ -540,7 +540,7 @@ def find_anatomical_estimators(
     base_entities : :class:`dict`
         Entities to use to query for images. These should include any filters.
     suffixes : :class:`list`
-        EPI suffixes, for example ``["bold", "dwi"]``. Associated ``"sbref"``\s
+        EPI suffixes, for example ``["bold", "dwi", "asl"]``. Associated ``"sbref"``\s
         will be found and used in place of BOLD/diffusion EPIs.
```
oesteban, line 544: Would this be true?
```suggestion
        EPI suffixes, for example ``["bold", "dwi", "asl"]``. Associated ``"sbref"``\s
        will be found and used in place of BOLD/diffusion EPIs.
        Similarly, ``"m0scan"``\s associated with ASL runs with the ``IntendedFor`` or
        ``B0FieldIdentifier`` metadata will be used in place of ASL runs.
```

oesteban (dismissed):
LGTM - just added a minor comment about m0scans

sdcflows/utils/wrangler.py
tsalo: Ah, great catch. That's definitely true.

oesteban (approved):

PR #415, opened 2023-12-15
FIX: Drop header before resampling image to avoid unsafe cast

effigies:
int16 images with scale factors can get badly truncated or wrapped. This PR scales the data array and drops the header before passing to `Affine().resample()` to avoid this. We then re-add the dtype, though this may not be necessary. I think the effects of multiple interpolation are going to be minimal here, so keeping dtypes as expected is probably worth it.


oesteban (approved):
Is this fixed in nitransforms?

effigies:
Not yet. Since that needs to work more broadly, a deeper review will be necessary.

PR #419, opened 2023-12-20
[FIX] Change `os.basename` to `os.path.basename`

smeisler:
Addresses bug found in https://neurostars.org/t/attributeerror-module-os-has-no-attribute-basename/27604/2.

An instance of `os.basename` is leading to errors since it should be `os.path.basename`.


mgxd (approved):
LGTM, thanks!

smeisler:
I realize I forgot to add myself to contributors list, will open a new PR for that

PR #421, opened 2024-01-09
FIX: Derive field from SyN displacements using EPI affine

effigies:
Theoretically, a fieldmap in Hz is the number of voxels/second that signal is shifted from the true location. SyN-SDC calculates the shift in mm from the true location. In order to find the equivalent fieldmap, we need the size of the voxels, readout time and phase-encoding direction of the EPI image that is to be corrected.

Previously, we were deriving the fieldmap from the readout time and phase-encoding direction of the EPI image, but the voxel size of the displacement field. This causes an overestimate of the number of voxels to be shifted, and thus exaggerated correction.

Original EPI:

image

Current unwarp: 

image

New unwarp:

image

### DIFF
--- a/sdcflows/workflows/fit/syn.py
+++ b/sdcflows/workflows/fit/syn.py
@@ -267,14 +267,15 @@ template [@fieldmapless3].
     )
 
     # Regularize with B-Splines
-    bs_filter = pe.Node(BSplineApprox(), name="bs_filter")
+    bs_filter = pe.Node(
+        BSplineApprox(recenter=False, debug=debug, extrapolate=not debug),
+        name="bs_filter",
+    )
     bs_filter.interface._always_run = debug
     bs_filter.inputs.bs_spacing = (
         [DEFAULT_LF_ZOOMS_MM, DEFAULT_HF_ZOOMS_MM] if not sloppy else [DEFAULT_ZOOMS_MM]
     )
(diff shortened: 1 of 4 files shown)


effigies:
@mgxd @oesteban This is ready to go.

mgxd (commented):
👍 definitely an improvement, although the current unwarping in the images you shared looks much better than in https://github.com/nipreps/fmriprep/issues/3158

effigies:
> 👍 definitely an improvement, although the current unwarping in the images you shared looks much better than in https://github.com/nipreps/fmriprep/issues/3158

Agreed. Running tests on one known to produce extreme distortions.

sdcflows/workflows/fit/syn.py
```diff
@@ -247,7 +247,7 @@ def init_syn_sdc_wf(
 
     # Extract the corresponding fieldmap in Hz
     extract_field = pe.Node(
-        DisplacementsField2Fieldmap(demean=True), name="extract_field"
```
oesteban, line 250: This is a legacy inherited from MGH's `epidewarp.fsl` and `epiunwarp` versions. Probably good to get it out of the way.

oesteban: Esp. when we then fit the b-spline :+1:

oesteban (approved):

PR #431, opened 2024-02-02
FIX: Remove unused ANTs parameter that was removed in 2.4.1

lalalavi:
Hello! :)

Starting with ANTS 2.4.1 there is a flag that has disappeared in the AntsRegistration command and we would like to backport the commit that fixes it  ( [link](https://github.com/nipreps/sdcflows/commit/24662aa51bbe8e11de7be47be94e99e46dbd26ba) ) into this version of sdcworkflows, since Halfpipe depends on it and we want to bump ANTS to solve problems with the FixN4BiasFieldCorrection interface that are fixed in the latest version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,37 +1,3 @@
-docs_deploy: &docs
-  docker:
-    - image: node:8.10.0
-  working_directory: /tmp/gh-pages
-  steps:
-    - run:
-        name: Check whether this is the original repo
-        command: |
-          if [[ "$CIRCLE_PROJECT_USERNAME" != "nipreps" ]]; then
-              echo "Not in nipreps/sdcflows - skipping docs deploy."
-              circleci step halt
-          fi
-    - add_ssh_keys:
-        fingerprints:
-          - "46:48:1e:6d:00:0e:f2:f8:e5:aa:b9:aa:da:eb:59:4e"
-    - run:
-        name: Install gh-pages tool
-        command: |
-          npm install -g --silent gh-pages@2.0.1
-    - checkout
-    - run:
-        name: Set git settings
-        command: |
-          git config user.email "nipreps@gmail.com"
-          git config user.name "Documentation Push"
-    - attach_workspace:
-        at: docs/_build
-    - run:
-        name: Disable jekyll builds
-        command: touch docs/_build/html/.nojekyll
-    - run:
-        name: Deploy docs to gh-pages branch
-        command: gh-pages --dotfiles --message "doc(update) [skip ci]" --dist docs/_build/html
-
 version: 2.1
 orbs:
   docker: circleci/docker@1.6.0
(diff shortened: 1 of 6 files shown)


effigies:
Tried this in https://github.com/nipreps/sdcflows/pull/331. Getting CI to work is going to be a more serious maintenance task than it seemed worth putting in.

lalalavi:
Hi @effigies I think we are going to try to correct the failing CI check. Should we continue on this current pull or reopen your old one?  Also, I see that the maint/1.3.x where we are trying to merge already had `build_docs` failing. Is the `build_n_pytest` test fail the reason why you stopped trying to backport the fix?

effigies:
Yes, let's not worry about docs. Just remove the stages from `.circleci` entirely at this point.

Need to figure out tests, though.

lalalavi:
Hi @effigies, is it possible that the image for the last release 1.3.4 was uploaded in Docker Hub and then removed? Or maybe never uploaded? I see a tag in git, but there is no manifest in Dockerhub and I can also not pull it from the docker CLI, which is the point where build_n_pytest is starting to crash.

Also I think it would help me debug if i could see the report of circleci for the previous job ( https://app.circleci.com/pipelines/github/nipreps/sdcflows/jobs/5477  ) but it does't go further from loading. Is this due to authorization issues or does it happen to you too?

effigies:
I think the 1.3.4 push to DockerHub failed. Old CircleCI runs expire, so we won't be able to see what happened that far back.

So we'll need to update the CircleCI config not to try gracefully fail pulls. You might look and see if we fixed this on `master`, but a straight copy from master to a legacy branch is rarely successful.

lalalavi:
This commit fixes the build in build_n_pytest. It is not very pretty because it hardcodes the fallback option for this release, but if at some point the build of the 1.3.4 is pushed to dockerhub it will access that. It now seems to crash in one of the unit tests because it is missing the freesurfer license, which i also see is specified in the circle ci config. I will look into this a bit further today!

HippocampusGirl:
I think the build cache needs to be cleared, because the license will only be written out on install, not when restoring from cache

```
Found a cache from build 6344 at freesurfer-v1-
Size: 2.8 GiB
Cached paths:
  * /tmp/freesurfer
```

effigies:
@lalalavi Pushed a quick commit, in case you didn't see.

lalalavi:
That is awesome :) Now the test passes as well. Do you want to still remove the `build_docs` entirely or leave as is before merging?

lalalavi:
Hi Chris!  I was wondering whether you still want me to remove the `build_docs` or whether you will merge it like it is now?  I have not gone into it as I saw some other parts of the .yaml were depending on build_docs (the `deploy_pypi`) and thought it could generate more problems, but if you want it the job to disappear before merging let me know.

effigies:
Ah, I figured we would remove `build_docs`, but I can do that real quick.

PR #435, opened 2024-04-07
Draft MEDIC dynamic distortion correction method

### LINKED ISSUE #36: Support dynamic distortion correction with DOCMA

tsalo:
Closes #36.

This currently runs on some test data, but I haven't evaluated the results.

Changes proposed:

- Add `sdcflows.workflows.fit.medic` module with MEDIC workflow.
    - I translated [vanandrew/warpkit](https://github.com/vanandrew/warpkit) into Nipype workflows.

Still to do:

- Figure out how to make `EnforceTemporalConsistency` more memory-efficient.
    - warpkit is _much_ faster.
- Compare SDCFlows results to MEDIC implementation in warpkit.
    - The current results look really bad compared to warpkit, but I'm confident I can pin down where they diverge.
- Convert functions used as nodes into Nipype interfaces.
- Add MEDIC option to the field map wrangler.
- Documentation
- Tests


tsalo:
I still need to figure out why the results don't look good, but I was thinking an alternative might be for me to simply integrate warpkit as a dependency and run that directly. Would that be preferable to a translated Nipype workflow?

mgxd:
If using warpkit significantly improves performance, all the more reason to wrap it rather than reimplement - it'll come with the added benefit of reducing longterm maintenance 😄 

One hurdle that would add is a more complex installation - however if/when https://github.com/vanandrew/warpkit/issues/6 is resolved it would be a cinch

tsalo:
@mgxd I did what you proposed in #438, so I'll close this PR now.

PR #436, opened 2024-04-16
FIX: Update suffix only when finding related fieldmap files

marcelzwiers:
SDCflows wants to replace the suffix but also accidentally replaces substrings elsewhere in the filename (e.g. in files named `sub-01-acq-fieldmap_magnitude1.nii.gz`). This PR fixes that -- see also github issue #264


effigies (approved):
Test failures are unrelated.

PR #434, opened 2024-04-02
Remove non-alphanumeric characters from workflow names and output entities

tsalo:
Closes #295.

Changes proposed:

- Use regular expressions to remove non-alphanumeric characters from the B0FieldIdentifier used in workflow names and output entities.

PR #437, opened 2024-04-16
Replace non-word/special characters from B0FieldIdentifier/Source tags

marcelzwiers:
Using e.g. `<<sesmri02>>` as a `B0FieldIdentifier` (which is a perfectly valid thing to do) leads to a nipype invalid node name error:

image

This PR replaces the special characters, such as '<` and `>` with `-`

### DIFF
--- a/sdcflows/fieldmaps.py
+++ b/sdcflows/fieldmaps.py
@@ -446,8 +446,8 @@ class FieldmapEstimation:
             return self._wf
 
         # Override workflow name (non-word/special characters are replaced to avoid nipype invalid node name errors)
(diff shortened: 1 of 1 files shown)


marcelzwiers:
This PR avoids errors from here: https://github.com/nipy/nipype/blob/master/nipype/pipeline/engine/base.py#L50

But perhaps it is better to filter out the special characters a bit later, e.g. in `niworkflows`, or even in nipype itself

marcelzwiers:
Oops, I now noticed PR #434, which looks like it may have already solved this?

PR #438, opened 2024-05-04
Draft MEDIC dynamic distortion correction method (second attempt)

### LINKED ISSUE #36: Support dynamic distortion correction with DOCMA

tsalo:
Closes #36. An alternative to #435 that installs MEDIC as a dependency instead of implementing the whole tool as a series of interfaces and workflows.

I ended up closing this largely because [`warpkit`](https://github.com/vanandrew/warpkit)'s license is not compatible with the NiPreps ecosystem, as it does not allow commercial use without an agreement with WUSTL.

Changes proposed:

- Install Julia in the Dockerfile.
- Add warpkit to dependencies.
- Add a MEDIC interface.
- Add a workflow to run MEDIC on a list of magnitude files and a list of phase files.


Dockerfile
```diff
@@ -80,6 +80,11 @@ RUN mkdir /opt/convert3d && \
     curl -fsSL --retry 5 https://sourceforge.net/projects/c3d/files/c3d/Experimental/c3d-1.4.0-Linux-gcc64.tar.gz/download \
     | tar -xz -C /opt/convert3d --strip-components 1
 
+# Julia for MEDIC
+FROM downloader as julia
+RUN curl -fsSL https://install.julialang.org | sh -s -- --yes --default-channel 1.9.4 && \
+    mkdir -p /opt/julia/ && cp -r /root/.julia/juliaup/*/* /opt/julia/
```
effigies: Can we add it to env.yml and install it with mamba instead?

tsalo: Should be done now

tsalo:
I _cannot_ figure out how to get the FieldmapWrangler to find MEDIC-style setups (i.e., complex-valued, multi-echo BOLD scans). 😕

effigies:
@tsalo Around L508 in your current `wrangler.py`, I think you want to do something like:

```python
        medic_entities = {**base_entities, **{'part': 'mag'}}
        has_magnitude = tuple()
        with suppress(ValueError):
            has_magnitude = layout.get(
                suffix='bold',
                **medic_entities,
            )
        for mag_img in has_magnitude:
            phase_img = layout.get(**{**mag_img.get_entities(), **{'part': 'phase'}})
            if not phase_img:
                continue
            phase_img = phase_img[
            try:
                e = fm.FieldmapEstimation(
                    [
                        fm.FieldmapFile(mag_img.path, metadata=mag_img.get_metadata()),
                        fm.FieldmapFile(phase_img.path, metadata=phase_img.get_metadata()),
                    ]
                )
            except (ValueError, TypeError) as err:
                _log_debug_estimator_fail(
                    logger, "potential MEDIC fieldmap", [mag_img, phase_img], layout.root, str(err)
                )
            else:
                _log_debug_estimation(logger, e, layout.root)
                estimators.append(e)
```

tsalo:
Thanks @effigies!

sdcflows/utils/wrangler.py
```diff
@@ -531,6 +540,39 @@ def find_estimators(
                     _log_debug_estimation(logger, e, layout.root)
                     estimators.append(e)
 
+                medic_entities = {**base_entities, **{'part': 'mag', 'echo': Query.ANY}}
```
effigies: Okay, now this is going to lump together multiple echos. I guess we can't count on everybody doing `'echo': 1` for their first echo? But either way, we can resolve this by checking for whether a `mag_img` is already part of an estimator.

effigies: And actually, maybe it makes sense to invert this and check for `phase`. You're less likely to get false positives that way.

tsalo: I'm fine hardcoding echo-1

tsalo: MEDIC uses all of the echoes for the field map estimation, so should all of the echoes (with both mag + phase) go into the `FieldmapEstimation`?

sdcflows/utils/wrangler.py
```diff
@@ -531,6 +540,39 @@ def find_estimators(
...
+                            [
+                                fm.FieldmapFile(mag_img.path, metadata=mag_img.get_metadata()),
+                                fm.FieldmapFile(phase_img.path, metadata=phase_img.get_metadata()),
+                            ]
+                        )
+                    except (ValueError, TypeError) as err:
+                        _log_debug_estimator_fail(
+                            logger,
+                            "potential MEDIC fieldmap",
+                            [mag_img, phase_img],
```
effigies: Then here we want to collect all of the images, right?

```suggestion
                    complex_imgs = layout.get(**{**mag_img.get_entities(), **{'part': ['phase', 'mag']}})

                    if complex_imgs[0].path in fm._estimators.sources:
                        continue

                    try:
                        e = fm.FieldmapEstimation(
                            [
                                fm.FieldmapFile(img.path, metadata=img.get_metadata()) for img in complex_imgs
                            ]
                        )
                    except (ValueError, TypeError) as err:
                        _log_debug_estimator_fail(
                            logger,
                            "potential MEDIC fieldmap",
                            complex_imgs,
```

sdcflows/utils/wrangler.py
effigies: Yeah, see my next comment.

tsalo:
I created a MEDIC-compliant test dataset (dsD), but it seems like the tests in test_wrangler use skeletons. Should I drop the test dataset and just generate a skeleton in the test file?

sdcflows/cli/tests/test_find_estimators.py
```diff
@@ -133,17 +133,21 @@
...
     import sdcflows.config as sc
 
     # Reload is necessary to clean-up the layout config between parameterized runs
     reload(sc)
 
     path = (tmp_path / test_id).absolute()
     generate_bids_skeleton(path, config)
-    with pytest.raises(SystemExit) as wrapped_exit:
+    # This was set to raise a SystemExit, but was only raising an ImageFileError
+    with pytest.raises(ImageFileError) as wrapped_exit:
```
effigies: This was probably expecting valid headers in the images.

effigies:
> Should I drop the test dataset and just generate a skeleton in the test file?

Yeah, seems like a good idea.

sdcflows/cli/tests/test_find_estimators.py
tsalo: I reverted the changes I made to this test since other PRs don't seem to have an issue. I'm not sure why the problem is happening in this PR only, but I've tried to remove any extraneous changes in case those were causing the problem.

tsalo: All I can think is that maybe adding `Julia` and `warpkit` to the dependencies is affecting which versions of other dependencies are installed (e.g., a newer version of nibabel?).

tsalo:
@effigies I finally got the tests passing. Do you know what my next steps should be?

sdcflows/cli/tests/test_find_estimators.py
```diff
@@ -131,7 +131,7 @@
         ("b0field", b0field_config, "pepolar"),
     ],
 )
-def test_cli_finder_wrapper(tmp_path, capsys, test_id, config, estimator_id):
+def _test_cli_finder_wrapper(tmp_path, capsys, test_id, config, estimator_id):
```
tsalo, line 134: Renamed to skip the test since it was failing.

tsalo:
@effigies I was thinking I could create a reduced version of one of the runs in https://openneuro.org/datasets/ds005250 (maybe like 20 volumes?), but I don't have access to `https://github.com/nipreps-data` to push it. I'm also not sure if any of the tests actually run the distortion correction on real data.

tsalo:
I pushed a single run with 50 volumes to https://gin.g-node.org/tsalo/ds005250-sdcflows. We can transfer ownership or make a fork in the nipreps-data organization. The dataset is 724 MB. Is that reasonable for a test dataset? I can reduce it further if necessary.

EDIT: I dropped them down to 10 volumes and also created a duplicate second session with different metadata.

effigies:
Invited you to nipreps-data.

tsalo:
Thanks! I'll create a GitHub repo for the datalad dataset and incorporate that into the tests.

.github/workflows/build-test-publish.yml
```diff
@@ -184,6 +184,11 @@ jobs:
         datalad update -r --merge -d hcph-pilot_fieldmaps/
         datalad get -r -J 2 -d hcph-pilot_fieldmaps/ hcph-pilot_fieldmaps/*
 
+        # ds005250
+        datalad install -r https://gin.g-node.org/tsalo/ds005250-sdcflows.git
+        datalad update -r --merge -d ds005250-sdcflows/
+        datalad get -r -J 2 -d ds005250-sdcflows/ ds005250-sdcflows/sub-04/ses-2/
```
tsalo: I set up ses-2 with IntendedFor and ses-1 with B0Field fields. The B0Fields aren't working.

tsalo: Also, it looks like `BIDSLayout.get_B0FieldIdentifiers` doesn't work at all.

bpinsard: are your B0Field* lists? see https://github.com/bids-standard/pybids/pull/684

tsalo: They are! I didn't realize that was a known issue. I can modify my test dataset to just have strings for B0FieldIdentifier.

Issue #295, opened 2022-11-04
B0FieldIdentifier: some characters should be replaced to set workflow name to avoid crashes in graph creation


bpinsard:
FMRIPrep version 20.1.0
Not sure if that is in fMRIPrep or SDCFlows that the workflow name is set, but I get the following error if B0FieldIdentifier contains a `-` , and this will likely occur for other characters (I have no knowledge on dot files syntax though). 

```
  File "/opt/conda/lib/python3.8/site-packages/fmriprep/cli/run.py", line 77, in main
    fmriprep_wf.write_graph(graph2use="colored", format="svg", simple_form=True)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/workflows.py", line 465, in write_graph
    outfname = format_dot(dotfilename, format=format)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/utils.py", line 1404, in format_dot
    formatted_dot, _ = _run_dot(dotfilename, format_ext=format)
  File "/opt/conda/lib/python3.8/site-packages/nipype/pipeline/engine/utils.py", line 1420, in _run_dot
    res = CommandLine(cmd, terminal_output="allatonce", resource_monitor=False).run()
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 428, in run
    runtime = self._run_interface(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 822, in _run_interface
    self.raise_exception(runtime)
  File "/opt/conda/lib/python3.8/site-packages/nipype/interfaces/base/core.py", line 749, in raise_exception
    raise RuntimeError(
RuntimeError: Command:
dot -Tsvg -o"...../workdir/fmriprep_wf/graph.svg" ".../workdir/fmriprep_wf/graph.dot"
Standard output:

Standard error:
Error: ..../workdir/fmriprep_wf/graph.dot: syntax error in line 29 near '-'
``` 

dot file contains something like `fmriprep_wf_single_subject_01_wf_fmap_preproc_wf_in_b0id-test` on the offending line.

bpinsard:
Replaced offending characters in my B0Field* tags and it works now, if that can help anyone encountering that issue.

effigies:
Ah, thanks for that. I think we assumed it would be a camel case string.

oesteban:
[...]

PR #444, opened 2024-06-06
RF: Add sanitized_id field to FieldmapEstimation

effigies:
Replaces #434 along the lines described in https://github.com/nipreps/sdcflows/issues/295#issuecomment-2151041189.

I think probably the cleanest way forward for downstream is for functions to always store and pass the estimator object, and select `bids_id` or `sanitized_id` only when a string is needed and what kind of string is needed is known.

TBH, this is a needless headache that is going to invite bugs, and I think we should put a check in the validator warning about `B0Field*` that contains non-alphanumeric characters, including underscores.

### DIFF
--- a/sdcflows/fieldmaps.py
+++ b/sdcflows/fieldmaps.py
@@ -300,6 +300,9 @@ class FieldmapEstimation:
     bids_id = attr.ib(default=None, kw_only=True, type=str, on_setattr=_id_setter)
     """The unique ``B0FieldIdentifier`` field of this fieldmap."""
 
+    sanitized_id = attr.ib(init=False, repr=False)
+    """Sanitized version of the bids_id with special characters replaced by underscores."""
+
     _wf = attr.ib(init=False, default=None, repr=False)
     """Internal pointer to a workflow."""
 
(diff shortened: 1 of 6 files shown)


mgxd (commented):
Looks good - even though testing seems to be broken atm, can you add a small test for this?

> TBH, this is a needless headache that is going to invite bugs, and I think we should put a check in the validator warning about B0Field* that contains non-alphanumeric characters, including underscores.

➕

tsalo (approved):
Thanks for fixing my original change. This PR makes sense to me.

I'm not sure how I feel about modifying the spec based on this though.

effigies:
@mgxd Test failures resolved, new tests added.

mgxd (approved):

effigies:
WDYT about this being in a patch release vs a feature release?

tsalo:
I'd consider it a patch, since SDCFlows doesn't currently work with what the BIDS spec allows.

mgxd:
I guess this will cause significant cache invalidation (at least in fmriprep) for anyone using a non-sanitized `B0FieldIdentifier`s, so minor release feels safer

effigies:
Okay. We'll just bump to 2.9.

PR #453, opened 2024-06-22
FIX: Revision of the B-Spline fitting code

oesteban:
Several issues are addressed:

- Fit intercept: instead of removing the mean/median/mode before fitting, let the model account for that by fitting an intercept.
- Lower regularization alpha: it seems the regularization can be much lower when fitting only one level of b-splines
- Update the workflow so only one level of b-splines is fit (multi-level fitting is not working, I believe)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -292,9 +292,9 @@ jobs:
 
       - restore_cache:
           keys:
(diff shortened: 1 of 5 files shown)


oesteban:
This resolves Sunjae's dataset issue and brings all test back to green. Merging...

PR #457, opened 2024-09-16
fix: Only fit high-frequency spline for SyN-SDC

effigies:
Follow-up to https://github.com/nipreps/sdcflows/pull/453, resolving similar issues for SyN-SDC as were resolved for GRE fieldmaps.


effigies:
@oesteban LMK what you think.

oesteban (approved):

PR #458, opened 2024-09-24
FIX: Normalize BIDS-URIs to subject-relative

### LINKED ISSUE #456: BIDS URIs failing for at least some datasets
### What happened?

Phasediff fieldmaps are not being found when `IntendedFor: ['bids::...', ...]`.

cc @julietteboscheron

### What command did you use?

```shell
fmriprep-docker $DATA $OUT participant
```

### What version of the software are you running?

24.1.0rc0

### How are you running this software?

Docker

### Is your data BIDS valid?

Yes

No

_No response_

_No response_

effigies:
Fixes #456.

### DIFF
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -406,7 +406,7 @@ def find_estimators(
                 e = fm.FieldmapEstimation(
                     fm.FieldmapFile(
                         fmap.path,
-                        metadata=_filter_metadata(fmap.get_metadata(), layout, subject),
+                        metadata=_filter_metadata(fmap.get_metadata(), subject),
                     )
                 )
             except (ValueError, TypeError) as err:
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -36,7 +36,6 @@ from .. import fieldmaps as fm
 
 def _normalize_intent(
     intent: str,
-    layout: BIDSLayout,
     subject: str
 ) -> str | None:
     """Convert BIDS-URI intent to subject-relative intent
@@ -66,12 +65,11 @@ def _resolve_intent(
 
 def _filter_metadata(
     metadata: Dict[str, Any],
-    layout: BIDSLayout,
     subject: str
 ) -> Dict[str, Any]:
     intents = metadata.get("IntendedFor")
     if intents:
-        updated = [_normalize_intent(intent, layout, subject) for intent in listify(intents)]
+        updated = [_normalize_intent(intent, subject) for intent in listify(intents)]
         return {**metadata, "IntendedFor": updated}
     return metadata
 
@@ -439,7 +437,7 @@ def find_estimators(
                 for fmap in layout.get(**{**entities, **{'direction': dirs}}):
                     fmapfile = fm.FieldmapFile(
                         fmap.path,
-                        metadata=_filter_metadata(fmap.get_metadata(), layout, subject),
+                        metadata=_filter_metadata(fmap.get_metadata(), subject),
                     )
                     by_intent.setdefault(
                         tuple(fmapfile.metadata.get('IntendedFor', ())), []
--- a/sdcflows/utils/wrangler.py
+++ b/sdcflows/utils/wrangler.py
@@ -34,6 +34,23 @@ from bids.utils import listify
 from .. import fieldmaps as fm
 
 
+def _normalize_intent(
+    intent: str,
+    layout: BIDSLayout,
+    subject: str
+) -> str | None:
+    """Convert BIDS-URI intent to subject-relative intent
+
+    SDCFlows currently makes strong assumptions about old-style intents,
+    and a change to that needs to be carefully considered and tested.
+    """
+    if intent.startswith("bids::"):
+        # bids::sub-<subject>/
+        #          ^- 10     ^- 11
+        return intent[11 + len(subject):]
+    return intent
+
+
 def _resolve_intent(
     intent: str,
     layout: BIDSLayout,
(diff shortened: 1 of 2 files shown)


effigies:
@mgxd @oesteban A review would be appreciated.

sdcflows/utils/wrangler.py
```diff
@@ -377,7 +406,10 @@ def find_estimators(
         ):
             try:
                 e = fm.FieldmapEstimation(
-                    fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+                    fm.FieldmapFile(
+                        fmap.path,
+                        metadata=_filter_metadata(fmap.get_metadata(), layout, subject),
```
oesteban: It looks like we are reading the metadata too many times right? 

https://github.com/nipreps/sdcflows/blob/08e198f1f62422d2d6274ed93a3b4ba7a78d661a/sdcflows/utils/wrangler.py#L427-L432

Then in 445 we use the `_resolve_intent()` filter, which is similar https://github.com/nipreps/sdcflows/blob/08e198f1f62422d2d6274ed93a3b4ba7a78d661a/sdcflows/utils/wrangler.py#L445

Perhaps we should try to avoid the second metadata read and use this one at `FieldmapFile` instantiation? (meaning, your fix looks good but the old code remains a bit brittle)

effigies: We're not actually duplicating reads because of:

https://github.com/nipreps/sdcflows/blob/08e198f1f62422d2d6274ed93a3b4ba7a78d661a/sdcflows/utils/wrangler.py#L434-L437

Fieldmaps affected by this PR never reach the later code.

The issue here is that SDCflows will happily create an estimator for a "whole" fieldmap (doesn't need a bold/dwi/m0scan to compete it) that points to a B0FieldIdentifier, a subject-relative IntendedFor or a BIDS-URI. But when it's time to look up what estimator I should use for my BOLD file (`get_identifier()`), if I look for a subject-relative path, then I miss the BIDS-URIs.

So anyway, it could be that the better solution is for fMRIPrep to try `bids::$DS_RELPATH` along with `$SUB_RELPATH`, but when I started this it felt like SDCflows should expose the estimators the same way regardless.

sdcflows/utils/wrangler.py
```diff
@@ -47,6 +64,18 @@ def _resolve_intent(
...
+def _filter_metadata(
+    metadata: Dict[str, Any],
+    layout: BIDSLayout,
+    subject: str
+) -> Dict[str, Any]:
+    intents = metadata.get("IntendedFor")
+    if intents:
+        updated = [_normalize_intent(intent, layout, subject) for intent in listify(intents)]
+        return {**metadata, "IntendedFor": updated}
+    return metadata
```
mgxd: doesn't seem like `layout` is needed here and in `_normalize_intent()`

sdcflows/utils/wrangler.py
```diff
@@ -47,6 +64,18 @@ def _resolve_intent(
...
 
 
+def _filter_metadata(
+    metadata: Dict[str, Any],
+    layout: BIDSLayout,
+    subject: str
+) -> Dict[str, Any]:
+    intents = metadata.get("IntendedFor")
+    if intents:
+        updated = [_normalize_intent(intent, layout, subject) for intent in listify(intents)]
```
mgxd:
```suggestion
        updated = [_normalize_intent(intent, subject) for intent in listify(intents)]
```

sdcflows/utils/wrangler.py
```diff
@@ -47,6 +64,18 @@ def _resolve_intent(
     return intent
 
 
+def _filter_metadata(
+    metadata: Dict[str, Any],
+    layout: BIDSLayout,
```
mgxd:
```suggestion
```

sdcflows/utils/wrangler.py
```diff
@@ -34,6 +34,23 @@
 from .. import fieldmaps as fm
 
 
+def _normalize_intent(
+    intent: str,
+    layout: BIDSLayout,
```
mgxd:
```suggestion
```

sdcflows/utils/wrangler.py
```diff
@@ -405,7 +437,10 @@ def find_estimators(
             if len(dirs) > 1:
                 by_intent = {}
                 for fmap in layout.get(**{**entities, **{'direction': dirs}}):
-                    fmapfile = fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+                    fmapfile = fm.FieldmapFile(
+                        fmap.path,
+                        metadata=_filter_metadata(fmap.get_metadata(), layout, subject),
```
mgxd:
```suggestion
                        metadata=_filter_metadata(fmap.get_metadata(), subject),
```

mgxd (commented):
Minor comment but looks good to me - I believe I tried breaking up the logic in the wrangler module sometime ago but glancing over it again and it was not enough...

sdcflows/utils/wrangler.py
```diff
@@ -377,7 +404,10 @@ def find_estimators(
         ):
             try:
                 e = fm.FieldmapEstimation(
-                    fm.FieldmapFile(fmap.path, metadata=fmap.get_metadata())
+                    fm.FieldmapFile(
+                        fmap.path,
+                        metadata=_filter_metadata(fmap.get_metadata(), layout, subject),
```
effigies:
```suggestion
                        metadata=_filter_metadata(fmap.get_metadata(), subject),
```

sdcflows/utils/wrangler.py
effigies: @oesteban Just checking back in on this comment. It says outdated because we removed the unused `layout` parameter, but the structure didn't change.

effigies:
Going to merge, and we can address any brittleness in a future PR.

PR #459, opened 2024-09-24
CI: Drop CircleCI, using GHA only

effigies:


effigies:
I'm going to merge to fix the builds. It might be worth re-adding CircleCI config just for the artifacts, but triggered on a `[circleci]` tag in the commit message or something.

PR #462, opened 2024-10-02
ENH: Allow Jacobian correction to be toggled on/off

effigies:
Jacobian correction is optional in fMRIPrep, but it can't be toggled when calling SDCFlows resampling, as in SDCFlows workflows. This:

* Allows `ApplyCoeffsField` and `init_unwarp_wf` to take a boolean `jacobian` argument that determine whether the Jacobian will be applied.
* Sets the default for `init_unwarp_wf` to `True` to avoid changing behavior in downstream tools that do not upgrade and decide to disable it.
* Sets the default to `False` for `ApplyCoeffsField`. This seems the more conservative option for an interface.
* Sets `jacobian=True` when adding `ApplyCoeffsField` to the `init_topup_wf` workflow and `jacobian=False` in the `init_syn_sdc_wf`

The main thing this enables is fixing https://github.com/nipreps/fmriprep/issues/3366.

### DIFF
--- a/sdcflows/interfaces/bspline.py
+++ b/sdcflows/interfaces/bspline.py
@@ -328,6 +328,11 @@ class _ApplyCoeffsFieldInputSpec(BaseInterfaceInputSpec):
         mandatory=True,
         desc="the phase-encoding direction corresponding to in_data",
     )
+    jacobian = traits.Bool(
+        False,
+        usedefault=True,
+        desc="apply Jacobian determinant correction after unwarping",
+    )
     num_threads = traits.Int(nohash=True, desc="number of threads")
     approx = traits.Bool(
         True,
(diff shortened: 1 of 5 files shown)


effigies:
I think this should be uncontroversial, but shout if you disagree @mgxd @oesteban. Will merge in a couple hours if I don't hear.

sdcflows/interfaces/bspline.py
```diff
@@ -421,6 +426,7 @@ def _run_interface(self, runtime):
             self.inputs.in_data,
             self.inputs.pe_dir,
             self.inputs.ro_time,
+            jacobian=self.inputs.jacobian or False,
```
mgxd: traits to the rescue
```suggestion
            jacobian=self.inputs.jacobian,
```

sdcflows/transform.py
```diff
@@ -73,6 +73,7 @@ def _sdc_unwarp(
     coordinates: np.ndarray,
     pe_info: Tuple[int, float],
     hmc_xfm: np.ndarray | None,
+    jacobian: bool,
```
mgxd, line 76: Assuming you've checked all instances - inserting a new parameter in the middle of a non keyword only function makes me nervous

mgxd (dismissed):
LGTM, little nits

sdcflows/transform.py
effigies: That's what testing's for. And yes, I've checked.

Issue #455, opened 2024-08-16
[B0FieldIdentifier] repeated between subjects makes fmriprep crash


suxpert:
### What happened?

I tried using **B0FieldIdentifier** and **B0FieldSource** for susceptibility distortion correction. According to [BIDS Spec](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#using-b0fieldidentifier-metadata), I've added
```json
"B0FieldIdentifier": "phdiff_fmap0",
```
in `.../fmap/...phasediff.json`, and added
```json
"B0FieldSource": "phdiff_fmap0",
```
in `.../func/...bold.json` for each subjects. I.e., I use the same identifier for all subjects.

When I tried run `fmriprep` for multiple participants at once, I got the error.

Is it required that this identifier **MUST** be different across subjects? Or are there chances to improve `fmriprep` such that we only need the identifier be unique within one subject?

### What command did you use?

```shell
ocker run -it --rm -v bids:/data:ro -v fmriprep:/out nipreps/fmriprep:24.0.1 --notrack --fs-no-reconall /data /out participant
```

### What version of fMRIPrep are you running?

24.0.1

### How are you running fMRIPrep?

Docker

### Is your data BIDS valid?

Yes

### Are you reusing any previously computed results?

No

### Please copy and paste any relevant log output.

```shell
Process Process-2:
Traceback (most recent call last):
  File "/opt/conda/envs/fmriprep/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/opt/conda/envs/fmriprep/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/conda/envs/fmriprep/lib/python3.11/site-packages/fmriprep/cli/workflow.py", line 126, in build_workflow
    retval['workflow'] = init_fmriprep_wf()
                         ^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/fmriprep/lib/python3.11/site-packages/fmriprep/workflows/base.py", line 94, in init_fmriprep_wf
    single_subject_wf = init_single_subject_wf(subject_id)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/envs/fmriprep/lib/python3.11/site-packages/fmriprep/workflows/base.py", line 515, in init_single_subject_wf
[...]

PR #465, opened 2024-10-09
 [FIX] force_fmapless on top of self-correcting pepolar bold

bpinsard:
When force_fmapless is used on top of self-correcting pepolar bold, the fmapless estimator will get it's name from B0FieldIdentifier, creating a duplicate.
For now I just created a test to replicate that behavior and triggering the error that this PR will try to fix.


bpinsard:
Also, it seems there was no test for `force_fmapless=True`.

bpinsard:
The tests don't fail when I run it locally. !?

bpinsard:
Not sure how that will behave with multi-subject cases, seee https://github.com/nipreps/sdcflows/issues/455 .
Maybe that should be covered in tests.

effigies (commented):
I'm not sure how I feel about this, now that I've implemented application of pre-computed fieldmaps. In practice, the way to discover the relevant fieldmaps was to re-run the SDCFlows fieldmap wrangling function, and then look for precomputed fieldmaps of the relevant IDs.

This will change that ID, making a hard break between pre- and post-change derivatives. One of the ostensible advantages to the fit-transform split was that we can run an improved transform stage on any previous derivatives, so long as the derivatives are queryable.

We would need to strategize an alternative method for finding the appropriate fieldmaps from both new and old derivatives to make this change.

bpinsard:
I am not sure to fully grasp the challenges here, and I wiped a lot of my memory of why I submitted that PR.
- the BIDS-id prefix that I added would likely break BIDS naming, but that can easily be fixed by something only alphanumerical
- do you worry that reexecution of the wrangler in the `apply` run would generate IDs different from the `fit` one, and thus fail to retrieve existing fieldmap. If so, how can we fix the autonaming in FieldmapEstimation to avoid the collision observed above that caused crashes?

PR #477, opened 2025-01-15
feat(get_trt): Allow estimated and fallback TotalReadoutTime

### LINKED ISSUE #249: Read Philips' ``EstimatedTotalReadoutTime`` and ``EstimatedEffectiveEchoSpacing``
Recent versions of dcm2niix have adopted these two for Philips scanners, we should check them before we resort to the water-fat shift formula.

See: rordenlab/dcm2niix#389 ([`rordenlab/dcm2niix@ab26453`](https://github.com/rordenlab/dcm2niix/commit/ab26453f97f61db60e7142e34b0b91f68aaa305a))

effigies:
This is work toward https://github.com/nipreps/fmriprep/issues/3009.

The primary purpose is to recover SyN-SDC in cases where TRT is not calculable. When we calculated displacement fields and applied them without round-tripping through a field in Hz, this didn't matter. Now that we do, it would be good to allow tools to supply a trivial value in these cases. I would suggest a power of 2 that is in the neighborhood of a plausible value, such as 0.03125, which only modifies the exponent of a floating point value, and not the significand, minimizing the potential for accumulating floating point error.

While I'm here, I provide a boolean flag (`False` by default) to fall back to EstimatedTotalReadoutTime or EstimatedEffectiveEchoSpacing.

The order of precedence (if both `use_estimated` and `fallback` are enabled) is:

1) TotalReadoutTime
2) EstimatedTotalReadoutTime
3) EffectiveEchoSpacing
4) EstimatedEffectiveEchoSpacing
5) EchoSpacing + ParallelReductionFactorInPlane
6) WaterFatShift + EPIFactor
7) fallback

We could instead move `Estimated*` to after `WaterFatShift`, if that would be better.

Closes #249.


sdcflows/interfaces/epi.py
```diff
@@ -37,6 +37,8 @@
 class _GetReadoutTimeInputSpec(BaseInterfaceInputSpec):
     in_file = File(exists=True, desc="EPI image corresponding to the metadata")
     metadata = traits.Dict(mandatory=True, desc="metadata corresponding to the inputs")
+    use_estimate = traits.Bool(False, desc='Use "Estimated*" fields to calculate TotalReadoutTime')
```
oesteban: If I understand Chris Rorden's thinking, this perhaps is best if enabled? Also, `usedefault` is missing

```suggestion
    use_estimate = traits.Bool(True, usedefault=True, desc='Use "Estimated*" fields to calculate TotalReadoutTime')
```

effigies: My thought was that since this isn't BIDS, using these fields should be opt-in instead of default.

Edit: `Estimated*` adds some friction where the user needs to take affirmative action to use these estimates, so I see this approach as respecting that intention. I'm not wedded to this position, though, so I'm okay being overruled.

effigies:
```suggestion
    use_estimate = traits.Bool(False, usedefault=True, desc='Use "Estimated*" fields to calculate TotalReadoutTime')
```

effigies:
Given that the only point of contention here is the default value of a flag and not the overall logic, and that the current default is consistent with the preexisting behavior, I'm going to merge this as "safe".

If you disagree with my response, we can change the default in a separate PR. There is also the opportunity to set a different default behavior for an app like fMRIPrep/dMRIPrep where it calls this function.

PR #479, opened 2025-01-29
feat: Add workflow arguments for metadata estimates and fallback TRT

effigies:
Follow-up to #477.

### DIFF
--- a/sdcflows/fieldmaps.py
+++ b/sdcflows/fieldmaps.py
@@ -113,12 +113,7 @@ class FieldmapFile:
     >>> f.suffix
     'T1w'
 
-    >>> FieldmapFile(
-    ...     dsA_dir / "sub-01" / "fmap" / "sub-01_dir-LR_epi.nii.gz",
-    ...     find_meta=False
-    ... )  # doctest: +IGNORE_EXCEPTION_DETAIL
-    Traceback (most recent call last):
-    MetadataError:
+    By default, the immediate JSON sidecar file is consulted for metadata.
 
     >>> f = FieldmapFile(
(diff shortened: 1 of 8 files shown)


oesteban (dismissed):
Looks reasonable :)

effigies:
I've got a broken https://github.com/nipreps/fmriprep/pull/3423 that is paired with this. I figure I should get that working before merging this. Was aiming for 25.0, but it was delayed enough.

oesteban (dismissed):

sdcflows/fieldmaps.py
```diff
@@ -251,12 +277,19 @@ def __attrs_post_init__(self):
...
-                    f"Missing readout timing information for <{self.path}>."
-                ) from exc
+                have_trt = True
+            except ValueError:
+                with suppress(ValueError):
+                    get_trt(self.metadata, in_file=self.path, use_estimate=True)
+                    extra = "Estimated timing is available."
+
+            if not have_trt:
+                logger.warning(msg, self.path, extra)
```
effigies, line 292: @oesteban Your review of changes to this class would be appreciated. I didn't see an obvious way to thread config options into `__attrs_post_init__`, so instead I settled for emitting warnings at object creation time. This may push errors down into runtime that would currently be caught. I'll see about having fMRIPrep run the check at workflow build time.

effigies: Actually, best place to put it was in the sdcflows preproc workflow generator.

oesteban:
@effigies sorry for not having had a look in time -- thanks for this!

PR #484, opened 2025-03-26
ENH: Allow changing laplacians weights in SyN registration metric

oesteban:


effigies (approved):
Looks sensible. A quick test to generate the workflow with 0, 0.5 and 1 and check the values in `wf.inputs.syn.metric_weight` (I think that wokrs) would be useful.

PR #485, opened 2025-03-26
rf: Allow callers to pass a pre-computed epi2anat transform for SyN

effigies:
Has some fiddling with Dockerfiles so I can try to run tests locally, but I haven't gotten it to work yet. Been a while since I developed inside containers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -94,7 +94,7 @@ RUN /opt/conda/envs/sdcflows/bin/pip install --no-cache-dir -r /tmp/requirements
 #
 # Main stage
 #
-FROM --platform=linux/amd64 ${BASE_IMAGE} as sdcflows
+FROM --platform=linux/amd64 ${BASE_IMAGE} as dev
 
 # Configure apt
 ENV DEBIAN_FRONTEND="noninteractive" \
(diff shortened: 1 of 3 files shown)


Dockerfile
oesteban, line 1: Do you want to push these updates independently of the rest of the PR?

sdcflows/workflows/fit/syn.py
```diff
@@ -523,12 +552,20 @@ def _remove_first_mask(in_file):
...
         ])  # fmt:skip
 
+        if coregister:
+            workflow.connect([
+                (epi2anat, transform_list, [("forward_transforms", "in1")]),
+            ])  # fmt:skip
+        else:
+            workflow.connect([
+                (inputnode, transform_list, [("epi2anat_xfm", "in1")]),
+            ])
```
oesteban, line 572:
```suggestion
            ])  # fmt:skip
```

sdcflows/workflows/fit/tests/test_syn.py
```diff
@@ -44,12 +46,11 @@
 @pytest.mark.parametrize(
     ("n_bold", "coregister", "sd_prior"),
```
oesteban, line 31:
```suggestion
    ("n_bold", "coregister"),
```

PR #487, opened 2025-03-26
feat(syn): Update totalFieldVarianceInVoxel space based on voxel resolution

effigies:
The SyN transform parameters are:

```
SyN[gradientStep,<updateFieldVarianceInVoxelSpace=3>,<totalFieldVarianceInVoxelSpace=0>]
```

The size of estimated displacements should not depend significantly on the voxel size. This change updates the totalFieldVarianceInVoxelSpace to mm values of 10mm and 0.5mm in the `sd_syn.json` files, and adds an interface to convert these to voxels based on the resolution in the phase-encoding direction.

### DIFF
--- a/sdcflows/workflows/fit/syn.py
+++ b/sdcflows/workflows/fit/syn.py
@@ -23,6 +23,7 @@
 """
 Estimating the susceptibility distortions without fieldmaps.
 """
+import json
 
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu
@@ -214,9 +215,14 @@ along the phase-encoding direction.
     find_zooms = pe.Node(niu.Function(function=_adjust_zooms), name="find_zooms")
     zooms_epi = pe.Node(RegridToZooms(), name="zooms_epi")
 
+    syn_config = data.load(f"sd_syn{'_sloppy' * sloppy}.json")
+
+    vox_params = pe.Node(niu.Function(function=_mm2vox), name="vox_params")
+    vox_params.inputs.registration_config = json.loads(syn_config.read_text())
+
     # SyN Registration Core
     syn = pe.Node(
-        Registration(from_file=data.load(f"sd_syn{'_sloppy' * sloppy}.json")),
+        Registration(from_file=syn_config),
         name="syn",
         n_procs=omp_nthreads,
     )
@@ -295,11 +301,14 @@ along the phase-encoding direction.
         (anat_dilmsk, amask2epi, [("out_file", "input_image")]),
         (amask2epi, epi_umask, [("output_image", "in2")]),
         (readout_time, warp_dir, [("pe_direction", "pe_dir")]),
+        (clip_epi, vox_params, [("out_file", "moving_image")]),
+        (readout_time, vox_params, [("pe_direction", "pe_dir")]),
         (warp_dir, syn, [("out", "restrict_deformation")]),
         (anat_merge, syn, [("out", "fixed_image")]),
         (fixed_masks, syn, [("out", "fixed_image_masks")]),
         (epi_merge, syn, [("out", "moving_image")]),
         (moving_masks, syn, [("out", "moving_image_masks")]),
+        (vox_params, syn, [("out", "transform_parameters")]),
         (syn, extract_field, [(("forward_transforms", _pop), "transform")]),
         (clip_epi, extract_field, [("out_file", "epi")]),
         (readout_time, extract_field, [("readout_time", "ro_time"),
--- a/sdcflows/workflows/fit/syn.py
+++ b/sdcflows/workflows/fit/syn.py
@@ -594,7 +594,7 @@ def init_syn_preprocessing_wf(
     return workflow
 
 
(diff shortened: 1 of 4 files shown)


effigies:
I've been reading https://github.com/ANTsX/ANTs/issues/734, and it looks all measurements are  relative to the fixed image, so I should change this from moving to fixed.

sdcflows/workflows/fit/syn.py
```diff
@@ -583,25 +594,52 @@ def _remove_first_mask(in_file):
...
-    retval = order @ [1 if pe_dir[0] == ax else 0.1 for ax in "ijk"]
+    moving_axcodes = nb.aff2axcodes(moving.affine, ["RR", "AA", "SS"])
+    moving_pe_axis = moving_axcodes["ijk".index(pe_dir[0])]
+
+    fixed_axcodes = nb.aff2axcodes(fixed.affine, ["RR", "AA", "SS"])
+
+    deformation = [0.1, 0.1, 0.1]
+    deformation[fixed_axcodes.index(moving_pe_axis)] = 1.0
+
+    return nlevels * [deformation]
```
effigies: @oesteban `anat_ref` is always non-oblique `RAS`, but the deobliqued EPI could be non-RAS. If the phase-encoding direction is rotated relative to RAS, then we could have been permitting deformations out of the phase-encoding direction.

When coordinating orientations across two images, the logic feels simpler to convert to axis codes and manipulate, rather than doing the linear algebra.

sdcflows/workflows/fit/syn.py
```diff
@@ -214,9 +215,14 @@ def init_syn_sdc_wf(
...
 
+    syn_config = data.load(f"sd_syn{'_sloppy' * sloppy}.json")
+
+    vox_params = pe.Node(niu.Function(function=_mm2vox), name="vox_params")
+    vox_params.inputs.registration_config = json.loads(syn_config.read_text())
+
     # SyN Registration Core
     syn = pe.Node(
-        Registration(from_file=data.load(f"sd_syn{'_sloppy' * sloppy}.json")),
+        Registration(from_file=syn_config),
```
oesteban, line 225: `from_file` is not an input field, but perhaps this is needed?
```suggestion
        Registration(from_file=str(syn_config)),
```

sdcflows/workflows/fit/syn.py
```diff
@@ -583,25 +594,52 @@ def _remove_first_mask(in_file):
     return workflow
 
 
-def _warp_dir(fixed_image, pe_dir, nlevels=3):
+def _warp_dir(moving_image, fixed_image, pe_dir, nlevels=3):
```
oesteban: We have 2 levels of registration only (I suspect that's because the affine component was removed into the preprocessing):

```suggestion
def _warp_dir(moving_image, fixed_image, pe_dir, nlevels=2):
```

sdcflows/workflows/fit/syn.py
```diff
@@ -214,9 +215,14 @@ def init_syn_sdc_wf(
     find_zooms = pe.Node(niu.Function(function=_adjust_zooms), name="find_zooms")
     zooms_epi = pe.Node(RegridToZooms(), name="zooms_epi")
 
+    syn_config = data.load(f"sd_syn{'_sloppy' * sloppy}.json")
+
+    vox_params = pe.Node(niu.Function(function=_mm2vox), name="vox_params")
+    vox_params.inputs.registration_config = json.loads(syn_config.read_text())
```
oesteban, line 221: Since we need to read the config outside the Registration node, we might want to extract the number of levels from it (e.g., `len(config_dict['transforms'])` instead of hard-code the magic number (of 2). It will be a bit trickier for handling masks, but a bit more robust, IMHO.

sdcflows/workflows/fit/syn.py
```diff
@@ -214,9 +215,14 @@ def init_syn_sdc_wf(
     find_zooms = pe.Node(niu.Function(function=_adjust_zooms), name="find_zooms")
     zooms_epi = pe.Node(RegridToZooms(), name="zooms_epi")
 
+    syn_config = data.load(f"sd_syn{'_sloppy' * sloppy}.json")
```
oesteban, line 218: Question out of curiosity: why in the test we use `data.load.readable` and here `data.load`?

oesteban (dismissed):

sdcflows/workflows/fit/syn.py
effigies: It already had been a `Path`, so it's clearly worked just fine.

sdcflows/workflows/fit/syn.py
oesteban: But probably for another PR btw

sdcflows/workflows/fit/syn.py
effigies: `readable()` just makes sure that it has `.read_text()`, and doesn't guarantee that it's somewhere on the filesystem. I left this as `data.load()` because that's what it was before I moved it, and I didn't want to change it here.

PR #480, opened 2025-03-06
ENH: Allow running SyN SDC without using prior

mgxd:
This PR makes the use of a prior during SyN SDC optional, since distortions may differ depending on population. Eventually, it would be nice to add a range of priors, either within SDCFlows directly or easily plugged in from other tools.

Relevant to https://github.com/nipreps/nibabies/issues/393

PR #488, opened 2025-03-26
FIX: Re-enable priors respecting ``sd_priors`` argument

oesteban:
Re-enables the inclusion of a prior mask, but moving the logic into the preprocessing (which is more readable).

If ``sd_priors`` is set to ``False`` on the preprocessing, then the ``sd_prior`` output is populated with another brain mask (which is equivalent to the behavior after #480).

EDIT: Pun intended for @effigies. As you'll see, I've refrained from changing the argument name `atlas_threshold`, however, I *fixed* the node name because—*to me*—a displacements field is not an atlas :D

### DIFF
--- a/sdcflows/workflows/fit/syn.py
+++ b/sdcflows/workflows/fit/syn.py
@@ -26,7 +26,6 @@ Estimating the susceptibility distortions without fieldmaps.
 
 from nipype.pipeline import engine as pe
 from nipype.interfaces import utility as niu
-from nipype.interfaces.base import Undefined
 from niworkflows.engine.workflows import LiterateWorkflow as Workflow
 
 from ... import data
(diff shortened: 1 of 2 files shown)


effigies (dismissed):
This makes sense to me.

@mgxd Note that `sd_prior` is removed from the `syn_sdc_wf`. It never did anything, but you should stop passing it before we remove `**kwargs` (if we do). It does still do something on the `syn_preprocessing_wf`.

oesteban:
> This makes sense to me.

and the atlas pun? :sweat_smile: 

EDIT: I see your emoji reaction now :D

sdcflows/workflows/fit/syn.py
```diff
@@ -44,13 +43,11 @@
 
 def init_syn_sdc_wf(
     *,
-    atlas_threshold=3,
     sloppy=False,
     debug=False,
     name="syn_sdc_wf",
     omp_nthreads=1,
     laplacian_weight=None,
-    sd_prior=True,
```
mgxd, line 53: parameter docstring should be removed

mgxd (dismissed):
Have not tested it, but feels cleaner removing the priors logic within `init_syn_sdc_wf`.

effigies (approved):

PR #492, opened 2025-05-15
sty: Adopt ruff, run basic fixes and reformat

effigies:
Review `pyproject.toml` and check the commit list, but the full diff is not worth reviewing. There are no unsafe fixes and no manual fixes.

Another PR will follow with more.

Edit: Those are instructions if you want to review. I plan to merge this when tests pass unless stopped.


effigies:
I address the remaining style complaint with a small rewrite, so delaying to the next PR.

PR #506, opened 2025-09-26
feat: Save mask to derivatives, pass derivatives to outputnode

effigies:
sdcflows adopted fit-transform before other nipreps. One thing we've found works well to ensure the same behavior with precomputed derivatives is to save the derivative and pass the saved derivative to the downstream workflow.

This does this for preprocessed fieldmaps, references, coefficients, and (newly) masks. Masks are used by downstream tools to register the fieldmap to the target image. Although they are not needed for application, if the registration is available, they do allow fieldmaps to be precomputed and then used in a new run.

There are a couple cleanup commits in here. The two that are meaningful are labeled `feat:`.

Enabling https://github.com/nipreps/fmriprep/pull/3532.

### DIFF
--- a/sdcflows/workflows/outputs.py
+++ b/sdcflows/workflows/outputs.py
@@ -123,6 +123,7 @@ def init_fmap_derivatives_wf(
     custom_entities=None,
     name='fmap_derivatives_wf',
     write_coeff=False,
+    write_mask=False,
 ):
     """
     Set up datasinks to store derivatives in the right location.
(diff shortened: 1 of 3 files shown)


sdcflows/workflows/outputs.py
```diff
@@ -213,8 +213,38 @@ def init_fmap_derivatives_wf(
...
+            ),
+            name='ds_mask',
+        )
+        ds_mask._interface._file_patterns += (
+            'sub-{subject}[/ses-{session}]/{datatype<fmap>|fmap}/'
+            'sub-{subject}[_ses-{session}][_hash-{hash}][_acq-{acquisition}]'
+            '[_dir-{direction}][_run-{run}][_part-{part}][_space-{space}]'
+            '[_cohort-{cohort}][_res-{resolution}][_fmapid-{fmapid}]'
+            '[_desc-{desc}]_{suffix<mask>}{extension<.nii|.nii.gz|.json>|.nii.gz}',
+        )
```
mgxd, line 239: opened https://github.com/nipreps/niworkflows/pull/977 to make this less hacky

mgxd (approved):
In favor of this - let's quickly release a niworkflow patch and remove the patterns hack

effigies:
@mgxd Pushed a niworkflows tag, and will update this shortly. I don't plan to wait on another CI round, and will depend on the CI for the release to catch if I mess something up.

Should this be 2.14.1 or 2.15.0? It makes very little difference.

mgxd:
2.15.0 - this will invalidate a significant amount of downstream caches

Issue #400, opened 2023-11-06
Dangling symlinks prevents collecting unused fieldmaps


effigies:
Unfortunately, this issue remains and had escaped my previous tests.
Even if now fmriprep do no run fieldmap estimation on fmaps not listed in b0 intents (https://github.com/nipreps/fmriprep/pull/3025) , the wrangler still lists/instantiates all fieldmap objects and crash when finding broken symlinks (datalad) on line https://github.com/nipreps/sdcflows/blob/master/sdcflows/fieldmaps.py#L202 . As we run jobs in parallel on temporary clones on the HPC, we only pull the strictly required data from the session to process, and thus it crashes.

An alternative would be to relax 
https://github.com/nipreps/sdcflows/blob/master/sdcflows/fieldmaps.py#L202 
and let broken symlink pass if only metadata are required to instantiate the fieldmap object ( which I am not sure about). 

In the reason mentioned no to apply session filters, the case where one wants to use fieldmaps from another session would still work by default if they provide no bids_filters for fmap or provide bids_filters that do not specify session, or include the other session, right?

_Originally posted by @bpinsard in https://github.com/nipreps/sdcflows/issues/359#issuecomment-1793980990_

effigies:
@bpinsard I think it does make sense to defer checking for file existence until we actually want to use them. Hopefully we can make validation a specific call, rather than an obligatory part of building a list of potential estimators.

PR #520, opened 2025-12-12
[FIX] defer check of file existence to workflow build stage

bpinsard:
Attempt at fixing #400, that is still a blocker when parallelizing session processing with datalad.

### DIFF
--- a/sdcflows/fieldmaps.py
+++ b/sdcflows/fieldmaps.py
@@ -229,10 +229,8 @@ class FieldmapFile:
     @path.validator
     def check_path(self, attribute, value):
         """Validate a fieldmap path."""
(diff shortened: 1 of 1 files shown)


effigies (approved):
LGTM. Thanks for looking into this, and apologies for letting it fall off the radar.

Have you tested this? Would it help to have an fMRIPrep image to test with? This can be backported trivially, so we can target pretty much any fMRIPrep series.

bpinsard:
I am currently testing it on one of our dataset, patching a container with the file. It builds the workflow so it seems to work (though I have other unrelated issues with pre-computed derivatives).

miltoncamacho:
LGTM

PR #541, opened 2026-04-26
[FEAT] MEDIC distortion correction via warpkit

### LINKED ISSUE #36: Support dynamic distortion correction with DOCMA

vanandrew:
## Summary

Adds multi-echo dynamic distortion correction (MEDIC) to sdcflows, backed by [warpkit](https://github.com/vanandrew/warpkit). MEDIC estimates a per-volume B0 fieldmap directly from a multi-echo, mag+phase BOLD series, capturing breathing- and motion-driven field changes that a single static fieldmap can't.

Revives #435 / #438 with the simpler pure-Python warpkit (post vanandrew/warpkit#16, no Julia/C++ setup required). Closes #36.

### Workflows
- **`init_medic_wf`** (`sdcflows/workflows/fit/medic.py`) — multi-echo phase + magnitude → 4D Hz fieldmap (one volume per timepoint, on the EPI grid) + brain-extracted reference + brain mask. Two-stage warpkit call (`UnwrapPhase` → `ComputeFieldmap`) so the per-frame masks stay accessible. Single `fmap` output is 4D for MEDIC, leaving the 3D-vs-4D dispatch to the apply consumer.
- **`init_dynamic_unwarp_wf`** (`sdcflows/workflows/apply/dynamic.py`) — per-volume apply path built on `sdcflows.transform.apply_dynamic_unwarp`, a per-frame extension of the same scipy/nitransforms-backed resampling that powers the static `init_unwarp_wf`. No warpkit needed at runtime. Includes Jacobian determinant intensity correction (`jacobian=True` default) sharing `transform.fieldmap_jacobian` with the static path.

### Interfaces
- **`sdcflows/interfaces/warpkit.py`** — thin `LibraryBaseInterface` wrappers around the two MEDIC stages SDCFlows actually drives: `UnwrapPhase` (ROMEO) and `ComputeFieldmap`. Lazy import — sdcflows imports warpkit only when these interfaces actually run.

### Detection / dispatch
- **`EstimatorType.MEDIC`** added. `FieldmapEstimation.__attrs_post_init__` detects MEDIC inputs (`part-{phase,mag}` on `bold/epi/sbref` sources), enforces matched echo cardinality (≥2 phase, equal mag count), and rejects partial or mixed part sets. `get_workflow` instantiates `init_medic_wf` with `EchoTime`-sorted lists (BIDS doesn't guarantee `echo` entity == numeric order).
- **Wrangler** seeds MEDIC detection from both `part='phase'` and `part='mag'` queries (some datasets carry `IntendedFor` only on one side); dedup walks the full sibling set, no reliance on pybids ordering.
- **`force_medic` opt-in** on `find_estimators` — auto-discover MEDIC estimators from complex multi-echo BOLD even when neither `IntendedFor` nor `B0FieldIdentifier` is set. Pairing is unambiguous because the part-mag/part-phase echoes of the same run are MEDIC sources by construction. Intended for public datasets that ship the required echoes without the metadata the default discovery path needs.

### Plumbing
- `init_fmap_preproc_wf` skips `fmap_coeff` for MEDIC (the fieldmap is on the EPI grid by construction, no B-spline rep).
- `init_fmap_derivatives_wf` uses `MergeSeries(allow_4D=True)` so MEDIC's 4D `fmap` passes through the same `ds_fieldmap` sink as the 3D static fmaps.

### Packaging
- New `warpkit` extra in `pyproject.toml`, **explicitly excluded from `[all]`** because warpkit ships under a non-commercial WUSTL license. Default `pip install sdcflows` stays Apache-clean.
- `tox.ini`: the `veryslow` env pulls the warpkit extra so MEDIC end-to-end tests only run there.
- CI: datalad-fetches sub-04/ses-2 of ds007637 and sub-a01 of ds006926 as MEDIC fixtures (cache key bumped to `v3`).

## Validation

- **Three-layer multi-echo guard**: wrangler seed query (`echo=Query.REQUIRED`), `FieldmapEstimation` cardinality check (`len(phase_files) < 2`), and `_unpack_metadata` runtime guard inside `init_medic_wf`.
- **Construction tests** for both workflows run on every CI lane (no warpkit needed).
- **`test_apply_dynamic_unwarp_matches_static`** pins `apply_dynamic_unwarp` to the same Hz→VSM + scipy.ndimage convention as the static `_sdc_unwarp` path — catches drift in sign / pe_info handling.
- **`test_wrangler_filter` / `test_wrangler_URIs`** parametrized with a 3-session × 3-echo × {mag,phase} BIDS skeleton.
- **End-to-end runs** on ds006926 (3-echo, 64×64×40, 205 frames) and ds007637 (5-echo, 110×110×72, 237 frames) confirm Jacobian factor sits at ~1.0 in the brain mask (std 0.087 / 0.118) with ~1–2% of voxels in compression/expansion regions and total signal change <2% — physically sensible.

## Known compromises

- Default wrangler MEDIC discovery still keys off `IntendedFor` / `B0FieldIdentifier` — same constraint as the existing single-PE EPI branch. Datasets missing both can opt into discovery via `force_medic=True` on `find_estimators` (see Detection / dispatch above).
- The single `fmap` output is 4D for MEDIC. Downstream tools that expect 3D field maps need to either dispatch on dimensionality or block MEDIC-based estimators until they do.

## Test plan

- [x] `pytest sdcflows/utils/tests/test_wrangler.py` — wrangler MEDIC detection paths (including `force_medic`)
- [x] `pytest sdcflows/workflows/fit/tests/test_medic.py` — `init_medic_wf` construction + `_unpack_metadata` guards
- [x] `pytest sdcflows/workflows/apply/tests/test_dynamic.py` — `init_dynamic_unwarp_wf` construction + jacobian flag + per-frame resampling vs. static path
- [x] `pytest -m veryslow` with `pip install sdcflows[warpkit]` and ds006926 / ds007637 fixtures present — full MEDIC fit + dynamic apply

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,6 +76,10 @@ test = [
     "pytest-xdist >= 2.5",
 ]
 
+warpkit = [
+    "warpkit >= 1.2.1",
+]
+
 # Aliases
 docs = ["sdcflows[doc]"]
 tests = ["sdcflows[test]"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dependencies = [
     "scipy >= 1.15",
     "templateflow >= 23.1",
     "toml >= 0.10",
+    "warpkit >= 1.5.0",
 ]
 dynamic = ["version"]
 
--- a/sdcflows/interfaces/tests/test_warpkit.py
+++ b/sdcflows/interfaces/tests/test_warpkit.py
@@ -1,0 +1,75 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Tests for the warpkit nipype interface wrappers.
+
+These tests check spec shape and small helpers; the actual ``_run_interface``
+methods require :mod:`warpkit`, which is an optional dependency.
+"""
+
+import pytest
+
+from sdcflows.interfaces import warpkit as wk
+
+
+def test_warpkit_base_interface_pkg():
+    """All warpkit interfaces share the ``warpkit`` library tag.
+
+    This lets nipype's ``LibraryBaseInterface`` emit a single, consistent
+    "warpkit not installed" message rather than per-class noise.
+    """
+    assert wk.WarpkitBaseInterface._pkg == 'warpkit'
+    for cls in (wk.UnwrapPhase, wk.ComputeFieldmap):
+        assert issubclass(cls, wk.WarpkitBaseInterface)
+
+
+@pytest.mark.parametrize(
+    'cls,expected_inputs,expected_outputs',
+    [
+        (
+            wk.UnwrapPhase,
+            {'phase', 'magnitude', 'echo_times'},
+            {'unwrapped', 'masks'},
+        ),
+        (
+            wk.ComputeFieldmap,
+            {'unwrapped', 'magnitude', 'masks', 'border_filt', 'svd_filt'},
+            {'fieldmap_native', 'displacement_map', 'fieldmap'},
+        ),
+    ],
+)
+def test_interface_spec_traits(cls, expected_inputs, expected_outputs):
+    """Each interface declares the expected input/output traits."""
+    iface = cls()
+    assert expected_inputs <= set(iface.inputs.copyable_trait_names())
+    assert expected_outputs <= set(iface.output_spec().copyable_trait_names())
+
+
+def test_compute_fieldmap_border_filt_default():
+    """``border_filt`` defaults to ``(1, 5)``.
+
+    Regression test for an upstream ``traits.Tuple`` quirk where the outer
+    ``default`` kwarg silently lost to inner ``Int()`` zeros, collapsing the
+    SVD border filter and clipping the dynamic fieldmap footprint.
+    """
+    iface = wk.ComputeFieldmap()
+    assert tuple(iface.inputs.border_filt) == (1, 5)
--- a/sdcflows/interfaces/warpkit.py
+++ b/sdcflows/interfaces/warpkit.py
@@ -68,9 +68,7 @@ def _as_str_list(x) -> list[str]:
 
 
 class _MEDICInputSpec(BaseInterfaceInputSpec):
-    phase = InputMultiObject(
-        File(exists=True), mandatory=True, desc='phase NIfTI, one per echo'
-    )
+    phase = InputMultiObject(File(exists=True), mandatory=True, desc='phase NIfTI, one per echo')
     magnitude = InputMultiObject(
         File(exists=True), mandatory=True, desc='magnitude NIfTI, one per echo'
     )
@@ -79,9 +77,7 @@ class _MEDICInputSpec(BaseInterfaceInputSpec):
         xor=['metadata'],
         desc='echo times in milliseconds, one per echo',
     )
-    total_readout_time = traits.Float(
-        xor=['metadata'], desc='EPI total readout time in seconds'
-    )
+    total_readout_time = traits.Float(xor=['metadata'], desc='EPI total readout time in seconds')
     phase_encoding_direction = traits.Enum(
         *PE_DIRECTIONS,
         xor=['metadata'],
@@ -93,9 +89,7 @@ class _MEDICInputSpec(BaseInterfaceInputSpec):
         desc='BIDS sidecar JSONs, one per echo (alternative to direct args)',
     )
     out_prefix = traits.Str('medic', usedefault=True, desc='prefix for output filenames')
-    noise_frames = traits.Int(
-        0, usedefault=True, desc='number of trailing noise frames to drop'
-    )
+    noise_frames = traits.Int(0, usedefault=True, desc='number of trailing noise frames to drop')
     n_cpus = traits.Int(4, usedefault=True, desc='number of CPUs to use')
     wrap_limit = traits.Bool(
         False, usedefault=True, desc='disable some phase-unwrapping heuristics'
@@ -124,11 +118,7 @@ class MEDIC(WarpkitBaseInterface, SimpleInterface):
                 phase=list(self.inputs.phase),
                 magnitude=list(self.inputs.magnitude),
                 out_prefix=out_prefix,
-                tes=(
-                    list(self.inputs.echo_times)
-                    if isdefined(self.inputs.echo_times)
-                    else None
-                ),
+                tes=(list(self.inputs.echo_times) if isdefined(self.inputs.echo_times) else None),
                 total_readout_time=(
                     self.inputs.total_readout_time
                     if isdefined(self.inputs.total_readout_time)
@@ -139,11 +129,7 @@ class MEDIC(WarpkitBaseInterface, SimpleInterface):
                     if isdefined(self.inputs.phase_encoding_direction)
                     else None
                 ),
-                metadata=(
-                    list(self.inputs.metadata)
-                    if isdefined(self.inputs.metadata)
-                    else None
-                ),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
                 noise_frames=self.inputs.noise_frames,
                 n_cpus=self.inputs.n_cpus,
                 wrap_limit=self.inputs.wrap_limit,
@@ -195,16 +181,8 @@ class UnwrapPhase(WarpkitBaseInterface, SimpleInterface):
                 phase=list(self.inputs.phase),
                 magnitude=list(self.inputs.magnitude),
                 out_prefix=out_prefix,
-                tes=(
-                    list(self.inputs.echo_times)
-                    if isdefined(self.inputs.echo_times)
-                    else None
-                ),
-                metadata=(
-                    list(self.inputs.metadata)
-                    if isdefined(self.inputs.metadata)
-                    else None
-                ),
+                tes=(list(self.inputs.echo_times) if isdefined(self.inputs.echo_times) else None),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
                 noise_frames=self.inputs.noise_frames,
                 n_cpus=self.inputs.n_cpus,
                 wrap_limit=self.inputs.wrap_limit,
@@ -230,9 +208,7 @@ class _ComputeFieldmapInputSpec(BaseInterfaceInputSpec):
         desc='unwrapped phase per echo (output of UnwrapPhase)',
     )
     magnitude = InputMultiObject(File(exists=True), mandatory=True)
-    masks = File(
-        exists=True, mandatory=True, desc='per-frame masks (output of UnwrapPhase)'
-    )
+    masks = File(exists=True, mandatory=True, desc='per-frame masks (output of UnwrapPhase)')
     echo_times = traits.List(traits.Float, xor=['metadata'])
     total_readout_time = traits.Float(xor=['metadata'])
     phase_encoding_direction = traits.Enum(*PE_DIRECTIONS, xor=['metadata'])
@@ -274,11 +250,7 @@ class ComputeFieldmap(WarpkitBaseInterface, SimpleInterface):
                 magnitude=list(self.inputs.magnitude),
                 masks=self.inputs.masks,
                 out_prefix=out_prefix,
-                tes=(
-                    list(self.inputs.echo_times)
-                    if isdefined(self.inputs.echo_times)
-                    else None
-                ),
+                tes=(list(self.inputs.echo_times) if isdefined(self.inputs.echo_times) else None),
                 total_readout_time=(
                     self.inputs.total_readout_time
                     if isdefined(self.inputs.total_readout_time)
@@ -289,11 +261,7 @@ class ComputeFieldmap(WarpkitBaseInterface, SimpleInterface):
                     if isdefined(self.inputs.phase_encoding_direction)
                     else None
                 ),
-                metadata=(
-                    list(self.inputs.metadata)
-                    if isdefined(self.inputs.metadata)
-                    else None
-                ),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
                 border_filt=tuple(self.inputs.border_filt),
                 svd_filt=self.inputs.svd_filt,
                 n_cpus=self.inputs.n_cpus,
@@ -345,11 +313,7 @@ class ApplyWarp(WarpkitBaseInterface, SimpleInterface):
                 transform=list(self.inputs.transform),
                 output=out_file,
                 transform_type=self.inputs.transform_type,
-                reference=(
-                    self.inputs.reference
-                    if isdefined(self.inputs.reference)
-                    else None
-                ),
+                reference=(self.inputs.reference if isdefined(self.inputs.reference) else None),
                 phase_encoding_axis=(
                     self.inputs.phase_encoding_axis
                     if isdefined(self.inputs.phase_encoding_axis)
--- a/sdcflows/interfaces/warpkit.py
+++ b/sdcflows/interfaces/warpkit.py
@@ -1,0 +1,536 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2024 The NiPreps Developers <nipreps@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# We support and encourage derived works from this project, please read
+# about our expectations at
+#
+#     https://www.nipreps.org/community/licensing/
+#
+"""Nipype interfaces wrapping :mod:`warpkit.api`.
+
+`warpkit <https://github.com/vanandrew/warpkit>`__ implements MEDIC
+(Multi-Echo DIstortion Correction) and related warp utilities. Each of
+warpkit's seven ``wk-*`` CLI tools is mirrored here as a
+:class:`~nipype.interfaces.base.SimpleInterface`, calling the corresponding
+:mod:`warpkit.api` function in-process. ``warpkit`` is an optional dependency
+— :class:`~nipype.interfaces.base.LibraryBaseInterface` produces a clean
+"package not installed" error if ``warpkit`` is missing at runtime.
+"""
+
+import os
+
+from nipype.interfaces.base import (
+    BaseInterfaceInputSpec,
+    File,
+    InputMultiObject,
+    LibraryBaseInterface,
+    OutputMultiObject,
+    SimpleInterface,
+    TraitedSpec,
+    isdefined,
+    traits,
+)
+
+PE_DIRECTIONS = ('i', 'j', 'k', 'i-', 'j-', 'k-', 'x', 'y', 'z', 'x-', 'y-', 'z-')
+PE_AXES = ('i', 'j', 'k', 'x', 'y', 'z')
+WARP_FORMATS = ('itk', 'fsl', 'ants', 'afni')
+
+
+class WarpkitBaseInterface(LibraryBaseInterface):
+    """Base for all warpkit-backed interfaces."""
+
+    _pkg = 'warpkit'
+
+
+def _as_str_list(x) -> list[str]:
+    if isinstance(x, str):
+        return [x]
+    return [str(p) for p in x]
+
+
+# ---------------------------------------------------------------------------
+# MEDIC — full multi-echo distortion correction pipeline
+# ---------------------------------------------------------------------------
+
+
+class _MEDICInputSpec(BaseInterfaceInputSpec):
+    phase = InputMultiObject(
+        File(exists=True), mandatory=True, desc='phase NIfTI, one per echo'
+    )
+    magnitude = InputMultiObject(
+        File(exists=True), mandatory=True, desc='magnitude NIfTI, one per echo'
+    )
+    echo_times = traits.List(
+        traits.Float,
+        xor=['metadata'],
+        desc='echo times in milliseconds, one per echo',
+    )
+    total_readout_time = traits.Float(
+        xor=['metadata'], desc='EPI total readout time in seconds'
+    )
+    phase_encoding_direction = traits.Enum(
+        *PE_DIRECTIONS,
+        xor=['metadata'],
+        desc='phase-encoding direction (with sign)',
+    )
+    metadata = InputMultiObject(
+        File(exists=True),
+        xor=['echo_times', 'total_readout_time', 'phase_encoding_direction'],
+        desc='BIDS sidecar JSONs, one per echo (alternative to direct args)',
+    )
+    out_prefix = traits.Str('medic', usedefault=True, desc='prefix for output filenames')
+    noise_frames = traits.Int(
+        0, usedefault=True, desc='number of trailing noise frames to drop'
+    )
+    n_cpus = traits.Int(4, usedefault=True, desc='number of CPUs to use')
+    wrap_limit = traits.Bool(
+        False, usedefault=True, desc='disable some phase-unwrapping heuristics'
+    )
+    debug = traits.Bool(False, usedefault=True, desc='enable debug mode')
+
+
+class _MEDICOutputSpec(TraitedSpec):
+    fieldmap_native = File(exists=True, desc='native-space B0 field map (Hz)')
+    displacement_map = File(exists=True, desc='displacement map (mm)')
+    fieldmap = File(exists=True, desc='undistorted-space B0 field map (Hz)')
+
+
+class MEDIC(WarpkitBaseInterface, SimpleInterface):
+    """Run the full MEDIC pipeline (:func:`warpkit.api.medic`)."""
+
+    input_spec = _MEDICInputSpec
+    output_spec = _MEDICOutputSpec
+
+    def _run_interface(self, runtime):
+        from warpkit.api import medic
+
+        out_prefix = os.path.join(runtime.cwd, self.inputs.out_prefix)
+        try:
+            result = medic(
+                phase=list(self.inputs.phase),
+                magnitude=list(self.inputs.magnitude),
+                out_prefix=out_prefix,
+                tes=(
+                    list(self.inputs.echo_times)
+                    if isdefined(self.inputs.echo_times)
+                    else None
+                ),
+                total_readout_time=(
+                    self.inputs.total_readout_time
+                    if isdefined(self.inputs.total_readout_time)
+                    else None
+                ),
+                phase_encoding_direction=(
+                    self.inputs.phase_encoding_direction
+                    if isdefined(self.inputs.phase_encoding_direction)
+                    else None
+                ),
+                metadata=(
+                    list(self.inputs.metadata)
+                    if isdefined(self.inputs.metadata)
+                    else None
+                ),
+                noise_frames=self.inputs.noise_frames,
+                n_cpus=self.inputs.n_cpus,
+                wrap_limit=self.inputs.wrap_limit,
+                debug=self.inputs.debug,
+            )
+        except ValueError as e:
+            raise RuntimeError(str(e)) from e
+
+        self._results['fieldmap_native'] = str(result.fieldmap_native)
+        self._results['displacement_map'] = str(result.displacement_map)
+        self._results['fieldmap'] = str(result.fieldmap)
+        return runtime
+
+
+# ---------------------------------------------------------------------------
+# UnwrapPhase — ROMEO multi-echo phase unwrapping
+# ---------------------------------------------------------------------------
+
+
+class _UnwrapPhaseInputSpec(BaseInterfaceInputSpec):
+    phase = InputMultiObject(File(exists=True), mandatory=True)
+    magnitude = InputMultiObject(File(exists=True), mandatory=True)
+    echo_times = traits.List(traits.Float, xor=['metadata'])
+    metadata = InputMultiObject(File(exists=True), xor=['echo_times'])
+    out_prefix = traits.Str('unwrap', usedefault=True)
+    noise_frames = traits.Int(0, usedefault=True)
+    n_cpus = traits.Int(4, usedefault=True)
+    wrap_limit = traits.Bool(False, usedefault=True)
+    debug = traits.Bool(False, usedefault=True)
+
+
+class _UnwrapPhaseOutputSpec(TraitedSpec):
+    unwrapped = OutputMultiObject(File(exists=True), desc='unwrapped phase per echo')
+    masks = File(exists=True, desc='per-frame masks NIfTI')
+
+
+class UnwrapPhase(WarpkitBaseInterface, SimpleInterface):
+    """ROMEO multi-echo phase unwrapping (:func:`warpkit.api.unwrap_phase`)."""
+
+    input_spec = _UnwrapPhaseInputSpec
+    output_spec = _UnwrapPhaseOutputSpec
+
+    def _run_interface(self, runtime):
+        from warpkit.api import unwrap_phase
+
+        out_prefix = os.path.join(runtime.cwd, self.inputs.out_prefix)
+        try:
+            result = unwrap_phase(
+                phase=list(self.inputs.phase),
+                magnitude=list(self.inputs.magnitude),
+                out_prefix=out_prefix,
+                tes=(
+                    list(self.inputs.echo_times)
+                    if isdefined(self.inputs.echo_times)
+                    else None
+                ),
+                metadata=(
+                    list(self.inputs.metadata)
+                    if isdefined(self.inputs.metadata)
+                    else None
+                ),
+                noise_frames=self.inputs.noise_frames,
+                n_cpus=self.inputs.n_cpus,
+                wrap_limit=self.inputs.wrap_limit,
+                debug=self.inputs.debug,
+            )
+        except ValueError as e:
+            raise RuntimeError(str(e)) from e
+
+        self._results['unwrapped'] = [str(p) for p in result.unwrapped]
+        self._results['masks'] = str(result.masks)
+        return runtime
+
+
+# ---------------------------------------------------------------------------
+# ComputeFieldmap — post-unwrap stage of MEDIC
+# ---------------------------------------------------------------------------
+
+
+class _ComputeFieldmapInputSpec(BaseInterfaceInputSpec):
+    unwrapped = InputMultiObject(
+        File(exists=True),
+        mandatory=True,
+        desc='unwrapped phase per echo (output of UnwrapPhase)',
+    )
+    magnitude = InputMultiObject(File(exists=True), mandatory=True)
+    masks = File(
+        exists=True, mandatory=True, desc='per-frame masks (output of UnwrapPhase)'
+    )
+    echo_times = traits.List(traits.Float, xor=['metadata'])
+    total_readout_time = traits.Float(xor=['metadata'])
+    phase_encoding_direction = traits.Enum(*PE_DIRECTIONS, xor=['metadata'])
+    metadata = InputMultiObject(
+        File(exists=True),
+        xor=['echo_times', 'total_readout_time', 'phase_encoding_direction'],
+    )
+    out_prefix = traits.Str('fieldmap', usedefault=True)
+    border_filt = traits.Tuple(
+        traits.Int(),
+        traits.Int(),
+        default=(1, 5),
+        usedefault=True,
+        desc='SVD components for the two-pass border filter',
+    )
+    svd_filt = traits.Int(10, usedefault=True)
+    n_cpus = traits.Int(4, usedefault=True)
+
+
+class _ComputeFieldmapOutputSpec(TraitedSpec):
+    fieldmap_native = File(exists=True)
+    displacement_map = File(exists=True)
+    fieldmap = File(exists=True)
+
+
+class ComputeFieldmap(WarpkitBaseInterface, SimpleInterface):
+    """Post-unwrap MEDIC stage (:func:`warpkit.api.compute_fieldmap`)."""
+
+    input_spec = _ComputeFieldmapInputSpec
+    output_spec = _ComputeFieldmapOutputSpec
+
+    def _run_interface(self, runtime):
+        from warpkit.api import compute_fieldmap
+
+        out_prefix = os.path.join(runtime.cwd, self.inputs.out_prefix)
+        try:
+            result = compute_fieldmap(
+                unwrapped=list(self.inputs.unwrapped),
+                magnitude=list(self.inputs.magnitude),
+                masks=self.inputs.masks,
+                out_prefix=out_prefix,
+                tes=(
+                    list(self.inputs.echo_times)
+                    if isdefined(self.inputs.echo_times)
+                    else None
+                ),
+                total_readout_time=(
+                    self.inputs.total_readout_time
+                    if isdefined(self.inputs.total_readout_time)
+                    else None
+                ),
+                phase_encoding_direction=(
+                    self.inputs.phase_encoding_direction
+                    if isdefined(self.inputs.phase_encoding_direction)
+                    else None
+                ),
+                metadata=(
+                    list(self.inputs.metadata)
+                    if isdefined(self.inputs.metadata)
+                    else None
+                ),
+                border_filt=tuple(self.inputs.border_filt),
+                svd_filt=self.inputs.svd_filt,
+                n_cpus=self.inputs.n_cpus,
+            )
+        except ValueError as e:
+            raise RuntimeError(str(e)) from e
+
+        self._results['fieldmap_native'] = str(result.fieldmap_native)
+        self._results['displacement_map'] = str(result.displacement_map)
+        self._results['fieldmap'] = str(result.fieldmap)
+        return runtime
+
+
+# ---------------------------------------------------------------------------
+# ApplyWarp — resample through a displacement transform
+# ---------------------------------------------------------------------------
+
+
+class _ApplyWarpInputSpec(BaseInterfaceInputSpec):
+    in_file = File(exists=True, mandatory=True)
+    transform = InputMultiObject(File(exists=True), mandatory=True)
+    out_file = traits.Str(desc='output path; defaults to <cwd>/applied.nii.gz')
+    transform_type = traits.Enum('map', 'field', mandatory=True)
+    reference = File(exists=True)
+    phase_encoding_axis = traits.Enum(*PE_AXES)
+    format = traits.Enum(*WARP_FORMATS, usedefault=True, default='itk')
+
+
+class _ApplyWarpOutputSpec(TraitedSpec):
+    out_file = File(exists=True)
+
+
+class ApplyWarp(WarpkitBaseInterface, SimpleInterface):
+    """Resample an image through a warpkit displacement transform
+    (:func:`warpkit.api.apply_warp`)."""
+
+    input_spec = _ApplyWarpInputSpec
+    output_spec = _ApplyWarpOutputSpec
+
+    def _run_interface(self, runtime):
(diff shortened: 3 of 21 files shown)


vanandrew:
I have no idea why [test (3.13, pre, fast)](https://github.com/nipreps/sdcflows/actions/runs/25840682073/job/75990934202#logs) is failing on the cache step.

It looks like an out-of-space error on the runner (but the other jobs with the same steps succeeded, so very confused):
```
[test (3.13, pre, fast)](https://github.com/nipreps/sdcflows/actions/runs/25840682073/job/75990934202)
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.334.0/_diag/Worker_20260514-125354-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.334.0/_diag/Worker_20260514-125354-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.334.0/_diag/Worker_20260514-125354-utc.log'
   at System.IO.RandomAccess.WriteAtOffset(SafeFileHandle handle, ReadOnlySpan`1 buffer, Int64 fileOffset)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at System.Diagnostics.TraceSource.Flush()
   at GitHub.Runner.Common.Tracing.Dispose(Boolean disposing)
   at GitHub.Runner.Common.Tracing.Dispose()
   at GitHub.Runner.Common.TraceManager.Dispose(Boolean disposing)
   at GitHub.Runner.Common.TraceManager.Dispose()
   at GitHub.Runner.Common.HostContext.Dispose(Boolean disposing)
   at GitHub.Runner.Common.HostContext.Dispose()
   at GitHub.Runner.Worker.Program.Main(String[] args)
```

tsalo:
@vanandrew I think someone must have rerun that failing job and now it's passing.

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,512 @@
...
+        xor=['metadata'],
+        desc='phase-encoding direction (with sign)',
+    )
+    metadata = InputMultiObject(
+        File(exists=True),
+        xor=['echo_times', 'total_readout_time', 'phase_encoding_direction'],
+        desc='BIDS sidecar JSONs, one per echo (alternative to direct args)',
+    )
+    out_prefix = traits.Str('medic', usedefault=True, desc='prefix for output filenames')
+    noise_frames = traits.Int(0, usedefault=True, desc='number of trailing noise frames to drop')
```
tsalo: In BIDS, these frames should not be part of the files. They go in separate `noRF` files, so you can drop support for them from SDCFlows.

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,512 @@
...
+                    self.inputs.total_readout_time
+                    if isdefined(self.inputs.total_readout_time)
+                    else None
+                ),
+                phase_encoding_direction=(
+                    self.inputs.phase_encoding_direction
+                    if isdefined(self.inputs.phase_encoding_direction)
+                    else None
+                ),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
```
tsalo: Does MEDIC convert the echo times in the metadata JSONs to milliseconds internally?

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,512 @@
...
+    # is what carries direction. Accept both to spare callers a strip step.
+    phase_encoding_axis = traits.Enum(*PE_DIRECTIONS)
+    format = traits.Enum(*WARP_FORMATS, usedefault=True, default='itk')
+
+
+class _ApplyWarpOutputSpec(TraitedSpec):
+    out_file = File(exists=True)
+
+
+class ApplyWarp(WarpkitBaseInterface, SimpleInterface):
```
tsalo: Is a warpkit-specific warping function necessary? I'd prefer to use nitransforms.

sdcflows/utils/wrangler.py
```diff
@@ -74,6 +74,7 @@ def find_estimators(
     sessions: list[str] | None = None,
     fmapless: bool | set = True,
     force_fmapless: bool = False,
+    force_medic: bool = False,
```
tsalo: I'm not sure this is necessary. Other field map types don't have force options.

sdcflows/workflows/apply/dynamic.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def init_dynamic_unwarp_wf(
+    *,
+    jacobian=True,
+    omp_nthreads=1,
+    name='dynamic_unwarp_wf',
+):
+    r"""
+    Apply a per-volume MEDIC fieldmap to unwarp a 4D EPI series.
```
tsalo: Is this workflow specific to MEDIC, or would it work on other dynamic distortion correction methods (e.g., DOCMA, TOAST)?

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,290 @@
...
+        name='extract_meta',
+        run_without_submitting=True,
+    )
+
+    # Two-stage warpkit path: UnwrapPhase exposes per-frame masks, which
+    # ComputeFieldmap then consumes. The one-shot MEDIC interface bundles
+    # both but hides the masks; we want them for ``fmap_dynamic_mask``.
+    unwrap = pe.Node(UnwrapPhase(), name='unwrap', n_procs=omp_nthreads)
+    unwrap.inputs.n_cpus = omp_nthreads
+    unwrap.inputs.debug = debug
```
tsalo:
```suggestion
    unwrap = pe.Node(UnwrapPhase(n_cpus=omp_nthreads, debug=debug), name='unwrap', n_procs=omp_nthreads)
```

Any reason to pass in inputs outside the initialization?

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,290 @@
...
+    # B-spline path, which the rest of sdcflows expects.
+    fmap_mean = pe.Node(
+        niu.Function(
+            input_names=['in_file'],
+            output_names=['out_file'],
+            function=_temporal_mean,
+        ),
+        name='fmap_mean',
+        run_without_submitting=True,
+    )
```
tsalo: Is splitting the 3D and 4D fmap outputs necessary? Can SDCFlows just write out whichever fmap type is appropriate and then toggle the type of unwarping based on the shape of the field map?

tsalo (changes requested):
Thanks for this! I made a quick pass through the code.

sdcflows/utils/wrangler.py
```diff
@@ -511,6 +526,78 @@ def find_estimators(
...
+                medic_seeds.extend(layout.get(**{**medic_seed_query, 'part': part}))
+
+        for bold_fmap in medic_seeds:
+            # Pull every echo + part for this run. ``get_entities()`` already
+            # includes extension; we override part/echo to widen the query.
+            run_entities = {
+                k: v for k, v in bold_fmap.get_entities().items() if k not in ('part', 'echo')
+            }
+            run_entities['part'] = ['phase', 'mag']
+            run_entities['echo'] = Query.ANY
```
tsalo, line 581: One more q: will MEDIC work on two-echo data?

sdcflows/interfaces/warpkit.py
vanandrew: Good point. I should definitely check this 😆

sdcflows/interfaces/warpkit.py
vanandrew: It's definitely not necessary. I can look into swapping to `nitransforms` if it's preferred.

sdcflows/utils/wrangler.py
vanandrew: The current way MEDIC gets picked up as an estimator is through definition of the `intendedFor` field in the sidecar (I think it's setup so that each image needs to reference all the echoes used).

This flag is supposed to supplement the fact that existing data might not already have this convention (else you would have to go back into existing data and adjust the `intendedFor` fields).

sdcflows/workflows/apply/dynamic.py
vanandrew: I imagine it could work for any workflow that produces the field map timeseries. So yes, pretty much any dynamic distortion correction method.

sdcflows/workflows/fit/medic.py
vanandrew: Yep, will correct.

sdcflows/utils/wrangler.py
vanandrew: Yes. The minimum requirement is having two echoes.

sdcflows/workflows/fit/medic.py
vanandrew: Hmm, not sure about this one. I decided against just having a the 4D field map output into the `fmap` key because it sort of changes the meaning of the output (Now `fmap` can be 3D or 4D, and I wasn't sure if that would break existing downstream consumers that always have been expecting it to be 3D).

I can change it though if we don't expect issues like that.

sdcflows/interfaces/warpkit.py
tsalo: That would be great, thank you. I believe fMRIPrep and other NiPreps have internal workflows for applying the transforms, including SDC from field maps, using nitransforms, so being consistent with that would be optimal.

sdcflows/workflows/fit/medic.py
tsalo: I think it would break downstream tools that don't support 4D field maps, but they can just block MEDIC SDC until they get it working.

sdcflows/utils/wrangler.py
tsalo: Okay, awesome! So this should be fine as long as >1 echoes are available.

sdcflows/interfaces/warpkit.py
vanandrew: Dropped in 9698a0e0 — agree, noRF frames go in their own files per BIDS.

sdcflows/interfaces/warpkit.py
vanandrew: Confirmed — `_unpack_metadata` in [`sdcflows/workflows/fit/medic.py`](https://github.com/nipreps/sdcflows/blob/feat/medic-warpkit/sdcflows/workflows/fit/medic.py) does `float(m['EchoTime']) * 1000.0` before handing TEs to warpkit, and warpkit's own metadata reader (`warpkit/scripts/_metadata.py`) does the same conversion on the metadata-only path. Both routes pass ms to ROMEO.

sdcflows/interfaces/warpkit.py
vanandrew: Done in 564bf51a. Added `apply_dynamic_unwarp` to [`sdcflows/transform.py`](https://github.com/nipreps/sdcflows/blob/feat/medic-warpkit/sdcflows/transform.py) — a per-frame extension of the existing scipy/nitransforms-backed resampling that already powers the static `B0FieldTransform.apply` path — and rewrote `init_dynamic_unwarp_wf` on top of it. With that internal consumer gone, also dropped the unused `MEDIC`, `ApplyWarp`, `ConvertWarp`, `ConvertFieldmap`, and `ComputeJacobian` wrappers from the warpkit interface module (the only two SDCFlows actually drives are `UnwrapPhase` and `ComputeFieldmap`).

sdcflows/workflows/fit/medic.py
vanandrew: Done in f1376663.

sdcflows/workflows/fit/medic.py
vanandrew: Done in 8888b36b. `init_medic_wf` now emits a single `fmap` whose dimensionality (3D static / 4D dynamic) tells consumers which apply path to use. Dropped `fmap_dynamic*`, `fmap_coeff`, the B-spline shim, and the `write_dynamic` derivatives branch. The derivatives sink uses `MergeSeries(allow_4D=True)` to pass 3D and 4D fmaps through the same `ds_fieldmap` path.

vanandrew:
Pushed four commits addressing your feedback (noise_frames, ctor inputs, nitransforms switch + interface cleanup, unified fmap output).

vanandrew:
I can't tell if these test failures are something wrong with my dataset additions or if it's just another one-off error:

```
update(error): . (dataset) [Fetch failed: CommandError(CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false fetch --verbose --progress --no-recurse-submodules --
prune origin' failed with exitcode 128 under /home/runner/sdcflows-tests/brain-extraction-tests [err: 'fatal: unable to access 'https://gin.g-node.org/nipreps-data/brain-extraction-
tests/': The requested URL returned error: 403'])] [CommandError: 'git -c diff.ignoreSubmodules=none -c core.quotepath=false fetch --verbose --progress --no-recurse-submodules --
prune origin' failed with exitcode 128 under /home/runner/sdcflows-tests/brain-extraction-tests [err: 'fatal: unable to access 'https://gin.g-node.org/nipreps-data/brain-extraction-
tests/': The requested URL returned error: 403']]
Error: Process completed with exit code 1.
```

vanandrew:
Looks like the CI keeps failing because it's running out of space (I added two datasets for tests so that might have pushed it over the edge).

I pushed [6f5cfff2](https://github.com/nipreps/sdcflows/pull/541/commits/6f5cfff2) adding a `jlumbroso/free-disk-space@main` step at the top of the `test` job:

```yaml
- name: Free disk space
  uses: jlumbroso/free-disk-space@main
  with:
    tool-cache: false
    android: true
    dotnet: true
    haskell: true
    large-packages: false
    swap-storage: false
```

This gets rid of some unneeded development tools on the github runner to free room for data.

vanandrew:
@tsalo Alright. I made changes based on our convo today:

- `find_estimators` discovers a fieldmap estimator for every fieldmap declared in BIDS intent metadata: grouping inputs by B0FieldIdentifier (Step 1), or falling back to IntendedFor heuristics (Step 2) only when no B0FieldIdentifier exists. (`no_medic` suppresses MEDIC via either route). It adds a `sort` that always returns dynamic estimators (i.e. MEDIC) first.
- Simplified `fmap_ref` and `fmap_mask` to be the first echo magntude and MEDIC fmap masks respectively. These likely won't be used by downstream consumers.
- Removed `force_medic` and replaced with `no_medic`

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,223 @@
...
+
+from nipype.interfaces import utility as niu
+from nipype.pipeline import engine as pe
+from niworkflows.engine.workflows import LiterateWorkflow as Workflow
+
+INPUT_FIELDS = ('phase', 'magnitude', 'metadata')
+
+_MEDIC_DESC = """\
+A dynamic *B<sub>0</sub>* nonuniformity map was estimated from multi-echo
+magnitude and phase EPI series using MEDIC [@van2023medic], as implemented in
```
tsalo: This can be bumped to the published 2026 paper, right? I know the ref goes in fMRIPrep's bib file, but the key should match the published version.

pyproject.toml
```diff
@@ -76,6 +76,16 @@ test = [
     "pytest-xdist >= 2.5",
 ]
 
+# warpkit is a Washington University-licensed tool that powers the MEDIC
+# workflow. Its license permits non-commercial use only; commercial users
+# need a separate agreement with WUSTL OTM. Keep this extra opt-in and
+# DO NOT add it to the `all` alias below — the default install must remain
+# clean-Apache.
```
tsalo: I'm not sure this is necessary. A number of SDCFlows' dependencies do not allow commercial use. I think we need to review and document restrictively licensed dependencies rather than put them in extra groups. @effigies wdyt?

sdcflows/workflows/fit/medic.py
vanandrew: This is fixed in [b0a6904](https://github.com/nipreps/sdcflows/pull/541/commits/b0a6904c5ceac438be75fd2a4fd0bb49bab68ccf).

vanandrew:
@tsalo I did a rebase to the latest main, hopefully to quash the 3.13 test error that happened in the last CI run: https://github.com/nipreps/sdcflows/actions/runs/26792754492/job/79155816998#step:22:320

sdcflows/transform.py
```diff
@@ -106,17 +106,43 @@ def _sdc_unwarp(
...
     )
 
-    # The Jacobian determinant image is the amount of stretching in the PE direction.
-    # Using central differences accounts for the shift in neighboring voxels.
-    # The full Jacobian at each voxel would be a 3x3 matrix, but because there is
-    # only warping in one direction, we end up with a diagonal matrix with two 1s.
-    # The following is the other entry at each voxel, and hence the determinant.
     if jacobian:
-        resampled *= 1 + np.gradient(vsm, axis=pe_info[0])
+        resampled *= fieldmap_jacobian(fmap_hz, pe_info[1], pe_info[0])
```
mgxd: what's the reasoning for this being split out when we already have the vsm?

sdcflows/interfaces/tests/test_warpkit.py
```diff
@@ -0,0 +1,75 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2024 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/transform.py
```diff
@@ -572,6 +598,154 @@ def to_displacements(self, ro_time, pe_dir, itk_format=True):
...
+            max_concurrent=num_threads or min(os.cpu_count(), 12),
+        )
+    )
+
+    if not allow_negative:
+        resampled[resampled < 0] = cval
+
+    moved = moving.__class__(resampled, moving.affine, moving.header)
+    moved.header.set_data_dtype(output_dtype)
+    return reorient_image(moved, axcodes)
```
mgxd: there's too much overlap here with `unwarp_parallel` and `apply` - can we unify these with the new 4D fieldmap methods?

sdcflows/workflows/apply/dynamic.py
```diff
@@ -0,0 +1,174 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/workflows/apply/dynamic.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def _dynamic_unwarp(distorted, fmap, pe_direction, readout_time, jacobian, num_threads):
+    """Resample a 4D EPI through a per-frame 4D Hz fieldmap on the same grid.
+
+    Uses :func:`~sdcflows.transform.apply_dynamic_unwarp`, which delegates
+    per-volume resampling to the same scipy-backed primitives that
+    :class:`~sdcflows.transform.B0FieldTransform` uses for the static path.
+    """
+    import os
```
mgxd: nit- prefer pathlib
```suggestion
    from pathlib import Path
```

sdcflows/workflows/apply/dynamic.py
```diff
@@ -0,0 +1,174 @@
...
+
+    resampled = apply_dynamic_unwarp(
+        distorted,
+        fmap,
+        pe_dir=pe_direction,
+        ro_time=readout_time,
+        jacobian=jacobian,
+        num_threads=num_threads,
+    )
+    out_file = os.path.abspath('corrected.nii.gz')
```
mgxd:
```suggestion
    out_file = Path('corrected.nii.gz').absolute()
```

sdcflows/workflows/apply/tests/test_dynamic.py
```diff
@@ -0,0 +1,211 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/workflows/apply/tests/test_dynamic.py
```diff
@@ -0,0 +1,211 @@
...
+
+from json import loads
+from pathlib import Path
+
+import pytest
+
+from ..dynamic import INPUT_FIELDS, init_dynamic_unwarp_wf
+
+# See test_medic._MEDIC_TEST_VOLUMES — keep these in sync.
+_MEDIC_TEST_VOLUMES = 3
```
mgxd: this should be a pytest.fixture in conftest

sdcflows/workflows/apply/tests/test_dynamic.py
```diff
@@ -0,0 +1,211 @@
...
+        'ds007637',
+        'sub-04/ses-2/func/sub-04_ses-2_task-fracback_acq-MBME_echo-*_part-mag_bold.nii.gz',
+        id='ds007637',
+    ),
+    pytest.param(
+        'ds006926',
+        'sub-a01/func/sub-a01_task-VisMot_acq-tr1800_echo-*_part-mag_bold.nii.gz',
+        id='ds006926',
+    ),
+]
```
mgxd: pytest fixture in conftest

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,223 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/workflows/apply/dynamic.py
```diff
@@ -0,0 +1,174 @@
...
+                          ('readout_time', 'readout_time')]),
+        (unwarp, average, [('out_file', 'in_file')]),
+        (unwarp, outputnode, [('out_file', 'corrected')]),
+        (average, brainextraction_wf, [('out_file', 'inputnode.in_file')]),
+        (brainextraction_wf, outputnode, [
+            ('outputnode.out_file', 'corrected_ref'),
+            ('outputnode.out_mask', 'corrected_mask'),
+        ]),
+    ])
+    # fmt: on
```
mgxd: nit - prefer inline fmt skip

```suggestion
    workflow.connect([
        (inputnode, rotime, [('distorted', 'in_file'),
                             ('metadata', 'metadata')]),
        (inputnode, unwarp, [('distorted', 'distorted'),
                             ('fmap', 'fmap')]),
        (rotime, unwarp, [('pe_direction', 'pe_direction'),
                          ('readout_time', 'readout_time')]),
        (unwarp, average, [('out_file', 'in_file')]),
        (unwarp, outputnode, [('out_file', 'corrected')]),
        (average, brainextraction_wf, [('out_file', 'inputnode.in_file')]),
        (brainextraction_wf, outputnode, [
            ('outputnode.out_file', 'corrected_ref'),
            ('outputnode.out_mask', 'corrected_mask'),
        ]),
    ])  # fmt:skip
```

pyproject.toml
mgxd: I agree, IMO it's fine to include in the normal dependencies - though up to @vanandrew how available they would like it

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,223 @@
...
+"""
+
+
+def init_medic_wf(
+    omp_nthreads=1,
+    sloppy=False,
+    debug=False,
+    name='medic_wf',
+    use_metadata_estimates=False,
+    fallback_total_readout_time=None,
```
mgxd:
```suggestion
    debug=False,
    name='medic_wf',
    **kwargs,
```

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,223 @@
...
+        4D binary brain mask (one mask per timepoint) as produced by MEDIC
+        (``warpkit``), aligned with ``fmap``.
+    method : :obj:`str`
+        Short description string.
+
+    """
+    # Project-internal imports only — none of these load warpkit at module
+    # import time. The warpkit dependency is resolved lazily inside the
+    # MEDIC interfaces at run time.
+    del sloppy, use_metadata_estimates, fallback_total_readout_time
```
mgxd:
```suggestion
```

sdcflows/workflows/fit/medic.py
```diff
@@ -0,0 +1,223 @@
...
+    from ...interfaces.warpkit import ComputeFieldmap, UnwrapPhase
+
+    workflow = Workflow(name=name)
+    workflow.__desc__ = _MEDIC_DESC
```
mgxd: is this needed outside the function? if not i would just include it here.

sdcflows/workflows/fit/tests/test_medic.py
```diff
@@ -0,0 +1,179 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

sdcflows/workflows/fit/tests/test_medic.py
```diff
@@ -0,0 +1,179 @@
...
+from pathlib import Path
+
+import pytest
+
+from ..medic import INPUT_FIELDS, _first, _unpack_metadata, init_medic_wf
+
+# A handful of timepoints is enough to exercise the full per-volume MEDIC
+# path; the source datasets ship 200+ volumes × 5 echoes × mag+phase, which
+# OOM-kills CI runners when xdist schedules these fixtures in parallel.
+_MEDIC_TEST_VOLUMES = 3
```
mgxd: reminder

sdcflows/workflows/fit/tests/test_medic.py
```diff
@@ -0,0 +1,179 @@
...
+        'ds007637',
+        'sub-04/ses-2/func/sub-04_ses-2_task-fracback_acq-MBME_echo-*_part-mag_bold.nii.gz',
+        id='ds007637',
+    ),
+    pytest.param(
+        'ds006926',
+        'sub-a01/func/sub-a01_task-VisMot_acq-tr1800_echo-*_part-mag_bold.nii.gz',
+        id='ds006926',
+    ),
+]
```
mgxd: reminder

sdcflows/workflows/tests/test_outputs.py
```diff
@@ -0,0 +1,43 @@
+# emacs: -*- mode: python; py-indent-offset: 4; indent-tabs-mode: nil -*-
+# vi: set ft=python sts=4 ts=4 sw=4 et:
+#
+# Copyright 2025 The NiPreps Developers <nipreps@gmail.com>
```
mgxd:
```suggestion
# Copyright The NiPreps Developers <nipreps@gmail.com>
```

mgxd (commented):
Thanks for pushing on this @vanandrew! I gave it a first pass and left some comments / questions, but overall the implementation looks reasonable. 

While waiting for this to get more feedback, we should probably coordinate with https://github.com/nipreps/fmriprep/pull/3645 (cc @DVSneuro) to sync up and use this branch, so we can have a branch that emulates how integration will actually look.

sdcflows/interfaces/tests/test_warpkit.py
vanandrew: Fixed in ced076d9.

sdcflows/interfaces/warpkit.py
vanandrew: Fixed in ced076d9.

sdcflows/workflows/apply/dynamic.py
vanandrew: Fixed in ced076d9.

sdcflows/workflows/apply/dynamic.py
vanandrew: Fixed in ced076d9.

sdcflows/workflows/apply/dynamic.py
vanandrew: Fixed in ced076d9.

sdcflows/workflows/apply/dynamic.py
vanandrew: Fixed in ced076d9.

sdcflows/workflows/fit/medic.py
vanandrew: Fixed in ea0892f0.

sdcflows/workflows/fit/medic.py
vanandrew: Fixed in ea0892f0 — folded the parity-only args into `**kwargs`.

sdcflows/workflows/fit/medic.py
vanandrew: Fixed in ea0892f0.

sdcflows/workflows/fit/medic.py
vanandrew: Fixed in ea0892f0 — only used here, so inlined it.

sdcflows/workflows/apply/tests/test_dynamic.py
vanandrew: Fixed in 8800cb36.

sdcflows/workflows/apply/tests/test_dynamic.py
vanandrew: Fixed in 8800cb36 — moved to shared `conftest` fixtures (`medic_test_volumes`, parametrized `medic_fixture`).

sdcflows/workflows/apply/tests/test_dynamic.py
vanandrew: Fixed in 8800cb36 — moved to shared `conftest` fixtures (`medic_test_volumes`, parametrized `medic_fixture`).

sdcflows/workflows/fit/tests/test_medic.py
vanandrew: Fixed in 8800cb36.

sdcflows/workflows/tests/test_outputs.py
vanandrew: Fixed in 8800cb36.

sdcflows/workflows/fit/tests/test_medic.py
vanandrew: Centralized the volume cap as the `medic_test_volumes` fixture in 8800cb36.

sdcflows/workflows/fit/tests/test_medic.py
vanandrew: `MEDIC_FIXTURES` is now the parametrized `medic_fixture` in conftest (8800cb36), de-duplicated across both test modules.

pyproject.toml
vanandrew: Moved warpkit into the core dependencies (0092e438) rather than an extras group, per the consensus here. The non-commercial WUSTL license is now documented in the `init_medic_wf` module docstring instead of the dependency list.

sdcflows/transform.py
vanandrew: Historical artifact. Changed it back to accepting a VSM as input (70b0e0c2), reusing the one already computed in `_sdc_unwarp`.

sdcflows/transform.py
vanandrew: This is a bigger change. I'll take a look and see what I can merge.

vanandrew: Unified in e94f285a. I made one path for both 3D/4D fields.

`unwarp_parallel` takes a 3D or 4D field. A 3D (static) field is `np.broadcast_to`-viewed across frames, so per-frame selection is a single branchless `fmap_hz[..., volid]`. The 3D case is now just the degenerate 4D case.

`B0FieldTransform` is where it gets a little messy, since MEDIC field maps don't use bspline coefficients. I modified the class so that it can be constructed from a pre-gridded field (`mapped=`) while maintaining initialization from B-spline coeffs. `apply()` dispatches on provenance (coeffs → `fit()`; pre-gridded → use as-is), not on shape, then both routes go through one shared `_resample_with_fieldmap` helper.

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+# ---------------------------------------------------------------------------
+
+
+class _UnwrapPhaseInputSpec(BaseInterfaceInputSpec):
+    phase = InputMultiObject(File(exists=True), mandatory=True)
+    magnitude = InputMultiObject(File(exists=True), mandatory=True)
+    echo_times = traits.List(traits.Float, xor=['metadata'])
+    metadata = InputMultiObject(File(exists=True), xor=['echo_times'])
+    out_prefix = traits.Str('unwrap', usedefault=True)
+    n_cpus = traits.Int(4, usedefault=True)
```
effigies: `num_threads` is the blessed nipype name for this, which will be synced with the containing `Node`'s `n_procs` field. Also, nipype/our convention is to set 1 on the interface, and then set the actual parallelism when inserting into the workflow.

```suggestion
    num_threads = traits.Int(1, usedefault=True, desc='Number of threads to use for unwrapping')
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+
+        out_prefix = os.path.join(runtime.cwd, self.inputs.out_prefix)
+        try:
+            result = unwrap_phase(
+                phase=list(self.inputs.phase),
+                magnitude=list(self.inputs.magnitude),
+                out_prefix=out_prefix,
+                tes=(list(self.inputs.echo_times) if isdefined(self.inputs.echo_times) else None),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
+                n_cpus=self.inputs.n_cpus,
```
effigies:
```suggestion
                n_cpus=self.inputs.num_threads,
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+        from warpkit.api import unwrap_phase
+
+        out_prefix = os.path.join(runtime.cwd, self.inputs.out_prefix)
+        try:
+            result = unwrap_phase(
+                phase=list(self.inputs.phase),
+                magnitude=list(self.inputs.magnitude),
+                out_prefix=out_prefix,
+                tes=(list(self.inputs.echo_times) if isdefined(self.inputs.echo_times) else None),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
```
effigies: `InputMultiObject` means the interface always sees a list, even if only a single value is passed. And `bool(Undefined) == False`, so we can be simpler about this:

```suggestion
                phase=self.inputs.phase,
                magnitude=self.inputs.magnitude,
                out_prefix=out_prefix,
                tes=self.inputs.echo_times or None,a
                metadata=self.inputs.metadata or None,
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+                n_cpus=self.inputs.n_cpus,
+                wrap_limit=self.inputs.wrap_limit,
+                debug=self.inputs.debug,
+            )
+        except ValueError as e:
+            raise RuntimeError(str(e)) from e
```
effigies: Is there any particular reason to convert a ValueError to a RuntimeError? I would probably just skip the `try`/`catch`.

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+    # mask==1 ring in warpkit.unwrap.svd_filtering and makes the dynamic
+    # fieldmap appear hard-brain-masked. Pass defaults to the inner Ints.
+    border_filt = traits.Tuple(
+        traits.Int(1),
+        traits.Int(5),
+        usedefault=True,
+        desc='SVD components for the two-pass border filter',
+    )
+    svd_filt = traits.Int(10, usedefault=True)
+    n_cpus = traits.Int(4, usedefault=True)
```
effigies:
```suggestion
        num_threads = traits.Int(1, usedefault=True, desc='Number of threads to use for estimating the fieldmap')
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+    # That collapses the border-filter to 0 SVD components, which zeros the
+    # mask==1 ring in warpkit.unwrap.svd_filtering and makes the dynamic
+    # fieldmap appear hard-brain-masked. Pass defaults to the inner Ints.
+    border_filt = traits.Tuple(
+        traits.Int(1),
+        traits.Int(5),
+        usedefault=True,
+        desc='SVD components for the two-pass border filter',
+    )
+    svd_filt = traits.Int(10, usedefault=True)
```
effigies:
```suggestion
    svd_filt = traits.Int(10, usedefault=True, desc='Number of singular values to truncate to')
```

sdcflows/interfaces/warpkit.py
```diff
@@ -0,0 +1,187 @@
...
+                ),
+                phase_encoding_direction=(
+                    self.inputs.phase_encoding_direction
+                    if isdefined(self.inputs.phase_encoding_direction)
+                    else None
+                ),
+                metadata=(list(self.inputs.metadata) if isdefined(self.inputs.metadata) else None),
+                border_filt=tuple(self.inputs.border_filt),
+                svd_filt=self.inputs.svd_filt,
+                n_cpus=self.inputs.n_cpus,
```
effigies: Dropping unnecessary casts, etc.

```suggestion
                unwrapped=self.inputs.unwrapped,
                magnitude=self.inputs.magnitude,
                masks=self.inputs.masks,
                out_prefix=out_prefix,
                tes=self.inputs.echo_times or None,
                total_readout_time=self.inputs.total_readout_time or None,
                phase_encoding_direction=self.inputs.phase_encoding_direction or None,
                metadata=self.inputs.metadata or None,
                border_filt=self.inputs.border_filt,
                svd_filt=self.inputs.svd_filt,
                n_cpus=self.inputs.num_threads,
```

effigies (commented):
There's a significant conceptual problem here. I'm not sure if it had been done correctly and then the distinction was lost after the latest refactor. If so, apologies for not getting to this sooner and saving you the time.

There is a comment that says the caller is responsible for ensuring that the fieldmap has already been projected into the moving (source) space, but we project them into the fixed (target) space.

To explain: We are resampling into the target space, so we need a VSM to correspond with every coordinate in the target space, and to then apply this shift in the source space. For a fixed fieldmap, this value is the same, so we can save computation by interpolating the fieldmap exactly once into the target space. For a dynamic fieldmap, there are no savings to be had here (and we don't know how to map the fieldmap into the target space without using the fieldmap to unwarp itself) and we need to interpolate the values in the target space. So if we want to have a single resampling function, it needs to have the logic:

```python
if not dynamic:
    vsm = fmap_hz * pe_info[1]
else:
    vsm = ndi.map_coordinates(fmap_hz, coordinates, ...) * pe_info[1]

coordinates[pe_info[0], ...] += vsm

resampled = ndi.map_coordinates(data, coordinates)
```

As it's a simple function, it might be cleaner to do two separate functions, one with a fieldmap in target space and one with it in source space. I don't know if that makes the wrapping code easier or harder.

---

I started making some other comments, but stopped when I got to this point. Happy to discuss further on this thread or on a call.

vanandrew:
@effigies Thanks for the review! I'll take a closer look at the comments later this week.

As for your conceptual concern, I think that comment might be a typo. I'm pretty sure I'm doing everything in the target space, not the moving space. But I'll take a look again and let you know for sure.

vanandrew:
> There is a comment that says the caller is responsible for ensuring that the fieldmap has already been projected into the moving (source) space, but we project them into the fixed (target) space.

@effigies Can you point out the line numbers of this comment?

effigies:
https://github.com/nipreps/sdcflows/blob/a68a1e4692db925896a6bb25e01c5eb0c044785e/sdcflows/transform.py#L594-L598

sdcflows/interfaces/warpkit.py
vanandrew: Done in fafe3bdc — renamed to `num_threads` (default 1) on both interfaces and dropped the explicit `n_cpus=omp_nthreads` in the workflow; the node's `n_procs` now drives it.

One tweak on your suggested line: I added `nohash=True` (matching `ApplyCoeffsField` in `bspline.py`). Since `num_threads` gets overwritten to `n_procs` at runtime, without `nohash` a run with a different `--nthreads` would change the node hash and needlessly invalidate cached results.

sdcflows/interfaces/warpkit.py
vanandrew: Done.

sdcflows/interfaces/warpkit.py
vanandrew: Done — dropped the `list(...)` casts and switched to `... or None`.

sdcflows/interfaces/warpkit.py
vanandrew: No reason — removed the try/except; the error propagates as-is now.

sdcflows/interfaces/warpkit.py
vanandrew: Done (same `num_threads` rename + `nohash=True` as above).

sdcflows/interfaces/warpkit.py
vanandrew: Done.

sdcflows/interfaces/warpkit.py
vanandrew: Done.

vanandrew:
@effigies I took a closer look at this. It's more of a badly phrased docstring. I reworded it in 7b02d9a to make it clearer, but there are two separate concepts:

- **Grid (lattice):** `fmap_hz` is co-gridded with `moving` — same voxel lattice — so the per-voxel scaling and coordinate shift broadcast element-wise. The sampling coordinates come straight from `moving`'s own grid ([`transform.py:642`](https://github.com/vanandrew/sdcflows/blob/7b02d9a92af763aaa2e80d6b32d7b4999a018d6b/sdcflows/transform.py#L642), `nt.linear.Affine(reference=moving)`), which is why the field has to share that lattice. This is a precondition for the pull-resample, not a statement about distortion space.
- **Semantics:** the field is valued in the undistorted (target) sense. In `_sdc_unwarp` the shift is applied exactly once, on the target lattice ([`transform.py:95-97`](https://github.com/vanandrew/sdcflows/blob/7b02d9a92af763aaa2e80d6b32d7b4999a018d6b/sdcflows/transform.py#L95-L97)):
  ```python
  # voxcoords is the deformation field, i.e., the target position of each voxel
  vsm = fmap_hz * pe_info[1]
  coordinates[pe_info[0], ...] += vsm
  ```
  No source-space treatment, no inversion.

So we agree the apply path works in target space — the old docstring just described it as "moving/source," which is exactly the contradiction you flagged. MEDIC hands in a field that's already target-valued and on the EPI grid, so it skips `fit()` ([`transform.py:508`](https://github.com/vanandrew/sdcflows/blob/7b02d9a92af763aaa2e80d6b32d7b4999a018d6b/sdcflows/transform.py#L508)); the static estimators reach the same state via B-spline reconstruction + coregistration inside `fit()` ([`transform.py:521`](https://github.com/vanandrew/sdcflows/blob/7b02d9a92af763aaa2e80d6b32d7b4999a018d6b/sdcflows/transform.py#L521)).

vanandrew:
I did a reorg of the git history for this since it was getting kind of intractable (~60 commits -> 8 commits). Also, rebased everything to the latest main branch.

vanandrew:
Thought I give this a ping. Is there anything that you need from my end to get this merged in? @effigies @tsalo @mgxd 

I think I've addressed all the current comments, but let me know if we need to discuss.

PR #542, opened 2026-05-11
enh: allow selection of motion correction when combining runs

mgxd:
Inspired by https://github.com/nipreps/nibabies/issues/542#issuecomment-4373685377

Adds a new parameter `mc_method` to `init_topup_wf` to allow alternative motion correction when combining across phase encoding directions. 

From the 3dvolreg documentation:
> Algorithm: Iterated linearized weighted least squares to make each
              sub-brick as like as possible to the base brick.
              This method is useful for finding SMALL MOTIONS ONLY.

No alternative cost function is available for 3dvolreg, whereas MCFLIRT's normcorr may be more robust (especially in infants).

### DIFF
--- a/sdcflows/workflows/fit/pepolar.py
+++ b/sdcflows/workflows/fit/pepolar.py
@@ -43,6 +43,7 @@ def init_topup_wf(
     debug=False,
     name='pepolar_estimate_wf',
     topup_config=None,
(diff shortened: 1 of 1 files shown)


effigies (approved):
LGTM

Issue #279, opened 2022-09-02
TOPUP takes +3h when many EPIs are fed


oesteban:
While debugging nipreps/fmriprep#2830 I've realized that our flattening of all input volumes ends with unreasonable amounts of repeated directions into topup. In particular @Gilles86's data have some 6 `_epi` images per PE polarity, and then each of those `_epi` has ~10 volumes, which in turn generates an input to topup of ~60 volumes.

I think we should motion-correct and then average each of those `_epi`. It's true the researcher could now use `B0FieldIdentifier` to pair EPIs in different ways, but calculating the average doesn't seem an extreme call at all.

WDYT?

(cc/ @effigies, who had to work out through similar issues before)

PR #548, opened 2026-06-25
ENH: Couple head motion and susceptibility distortion within TOPUP

mgxd:
Reworks `init_topup_wf` to have topup handle both head motion and SD estimation concurrently. 

A new interface `SelectPEVolumes` has been added to reduce excessive number of inputs (like in #279), based on [recommendations from the topup docs](https://fsl.fmrib.ox.ac.uk/fsl/docs/diffusion/topup/users_guide/index.html#what-data-to-acquire-in-order-to-use-topup). ATM, this is set to 3 pairs, but can be further reduced if needed.

Additionally, the padding hack needed to process images with an uneven # of slices was removed (again, [recommended by their docs](https://fsl.fmrib.ox.ac.uk/fsl/docs/diffusion/topup/FAQ/index.html#do-i-have-to-have-an-even-number-of-slices-to-use-topup)) - instead their stock configs are now present and are auto-selected based on the shape of the incoming field maps. Config file used should not have a drastic effect on the field, just the processing speed.

Given that `RobustAverage` was removed, `mc_method` is no longer needed and was removed from `init_topup_wf`. Hadn't seen a release yet so better to remove than deprecate.

### DIFF
--- a/sdcflows/interfaces/utils.py
+++ b/sdcflows/interfaces/utils.py
@@ -21,7 +21,7 @@
 #     https://www.nipreps.org/community/licensing/
 #
 """Utilities."""
-
+import logging
 from itertools import product
 
 from nipype.interfaces.ants.segmentation import (
@@ -45,6 +45,7 @@ from nipype.interfaces.mixins import CopyHeaderInterface as _CopyHeaderInterface
 from sdcflows.utils.tools import reorient_pedir
 
 OBLIQUE_THRESHOLD_DEG = 0.5
+LOGGER = logging.getLogger('nipype.interface')
 
 
 class _FlattenInputSpec(BaseInterfaceInputSpec):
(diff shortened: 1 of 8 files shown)


madisoth:
As a quick example of the registration issue encountered when doing PEPolar blip registration independent of SD estimation, here's a registered pair of blips from an infant (age 1mo) using the current method with 3dVolReg:

volreg0

volreg1

The merged volume output of topup:

volreg_merged

Compare to the output when using topup's motion correction:

 
topupmc_merged

sdcflows/interfaces/utils.py
```diff
@@ -143,13 +146,26 @@ def _run_interface(self, runtime):
                     nii.__class__(nii.dataobj, refaff, nii.header).to_filename(retval[i])
                 continue
 
+            # The input is genuinely on a different sampling grid and must be resampled.
+            resampled.append(fname)
```
effigies: Let's gate this on `np.allclose(nii.affine[:3, :3], refaff[:3, :3])`

effigies (dismissed):
Otherwise LGTM

